PR #1, opened 2016-03-31
Bhargav/feature/robotofontimpl

bhargav1:


app/src/main/java/nu/yona/app/customview/FontCache.java
```diff
@@ -0,0 +1,35 @@
...
+import android.content.Context;
+import android.graphics.Typeface;
+
+import java.util.HashMap;
+
+public class FontCache {
+
+    private static HashMap<String, Typeface> fontCache = new HashMap<>();
+
+    public static Typeface getTypeface(String fontname, Context context) {
```
vasakinnar: @bhargav1 Why static?

PR #23, opened 2016-04-12
APPDEV-31

vasakinnar:
API implementation for resend confirmation code

Signed-off-by: Kinnar Vasa <kvasa@mobiquityinc.com>


app/src/main/java/nu/yona/app/api/manager/network/AuthenticateNetworkImpl.java
```diff
@@ -83,4 +83,32 @@ public void onFailure(Call<User> call, Throwable t) {
...
+                    }
+                }
+
+                @Override
+                public void onFailure(Call call, Throwable t) {
+                    listener.onError(new ErrorMessage(t.getMessage()));
+                }
+            });
+        } catch (Exception e) {
+
```
bhargav1: Why we not handling exception here ?

PR #40, opened 2016-04-15
Implemented API for Categories and Goals

bhargav1:


app/src/main/java/nu/yona/app/api/manager/dao/ActivityCategoriesDAO.java
```diff
@@ -10,69 +10,64 @@
...
+import nu.yona.app.api.model.ActivityCategories;
 import nu.yona.app.listener.DataLoadListener;
-import nu.yona.app.listener.DataLoader;
 
 /**
  * Created by kinnarvasa on 14/04/16.
  */
 public class ActivityCategoriesDAO extends BaseDAO {
 
+    private final String ID = "1";
```
vasakinnar: This should not be here as well, values.put(DBConstant.ID, ID); should store incremental value or there is no require of ID

bhargav1: we are using this for updating that Category object into database , as what we are doing same for User.

PR #41, opened 2016-04-15
Implementation/bhargavsuthar/appdev 64

bhargav1:


app/src/main/java/nu/yona/app/customview/YonaFontUtils.java
```diff
@@ -50,13 +45,16 @@ private static Typeface selectTypeface(Context context, int textStyle) {
             http://developer.android.com/reference/android/R.styleable.html#TextView_textStyle
             */
         switch (textStyle) {
-            case Typeface.BOLD: // bold
+            case 12: // bold
```
vasakinnar: @bhargav1 It should be from integer constant/xml.

PR #48, opened 2016-04-20
[APPDEV-65] Handle the back press view

bhargav1:


app/src/main/java/nu/yona/app/ui/challenges/ChallengesFragment.java
```diff
@@ -102,4 +111,25 @@ private void updateChallengeCounter(TabLayout.Tab tab, int count) {
         //Todo update the count of goal set
     }
 
+    public boolean isChildViewVisible() {
+        if (tabLayout.getSelectedTabPosition() == 0) {
```
vasakinnar: @bhargav1 can't we use switch here with TAB_INDEX... variables?

PR #50, opened 2016-04-20
Feature/bhargavsuthar/appdev 65

bhargav1:


app/src/main/java/nu/yona/app/ui/challenges/BaseGoalCreateFragment.java
```diff
@@ -104,9 +113,14 @@ private synchronized void filterCategoriesGoal() {
         if (userGoals != null && userGoals.getEmbedded() != null && userGoals.getEmbedded().getYonaGoals().size() > 0) {
             for (YonaGoal mYonaGoal : userGoals.getEmbedded().getYonaGoals()) {
                 if (mYonaGoal != null) {
-                    if (mYonaGoal.getType().equalsIgnoreCase(GoalsEnum.BUDGET_GOAL.getActionString())) {
+                    for (Map.Entry<String, String> entry : mGoalCategoriesMap.entrySet()) {
+                        if (entry.getValue().equals(mYonaGoal.getLinks().getYonaActivityCategory().getHref())) {
+                            mYonaGoal.setActivityCategoryName(entry.getKey());
```
vasakinnar: @bhargav1  You can add break inside if once value match, so it don't complete entire loop if object found before end.

PR #101, opened 2016-05-06
Checked Null on dataloader listener

bhargav1:


vasakinnar:
This is no more valid and no further require.

PR #146, opened 2016-05-24
[APPDEV-314] When user clicks on Challenge Opslaan Button in Tegoed w…

bhargav1:
…ithout selecting time user is not shown message as user does not know what to do.


app/src/main/res/values-nl/strings.xml
```diff
@@ -112,7 +112,7 @@
     <string name="timepickerendtimelable">Select time till you want to spent:</string>
     <string name="timepickererror">Current time must be future time.</string>
     <string name="timepickerselect">Select time you want to spent:</string>
-    <string name="add_timezone">Please add atlest one time to add in timezone.</string>
+    <string name="add_goal_error">Please add atlest one time to add in %1$s.</string>
```
vasakinnar: @bhargav1 please update this string in translation.xls file.

PR #218, opened 2016-06-16
Revert "Fixed : after user blocked user not able to see the block message"

bhargav1:
Reverts yonadev/yona-app-android#217


vasakinnar:
@bhargav1 Already merged!

PR #464, opened 2017-09-15
Feature/mct 79

imspatni:
Added feature support to display NoGo Notification URL & Event Time.

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/dashboard/CustomPageAdapter.java
+++ b/app/src/main/java/nu/yona/app/ui/dashboard/CustomPageAdapter.java
@@ -21,11 +21,13 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import nu.yona.app.R;
 import nu.yona.app.YonaApplication;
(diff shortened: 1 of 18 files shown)


app/src/main/java/nu/yona/app/ui/dashboard/CustomPageAdapter.java
```diff
@@ -106,6 +121,25 @@ private ViewGroup initiateDayActivityReport(ViewGroup collection, int position)
         goalType = (YonaFontTextView) spreadView.findViewById(R.id.goalType);
         goalScore = (YonaFontTextView) spreadView.findViewById(R.id.goalScore);
         mSpreadGraph = (SpreadGraph) spreadView.findViewById(R.id.spreadGraph);
+
+        // This we have considered as an List of link will be display to user as per displaying in
```
punitsharma: can't we declare all the List and layout manager previously and just notify adapter from here. 
as this logic will get called each time unnecessarily.

imspatni: This must be execute for each ViewPager page it load. Other thing is this is also the place where all view child get initialised and return as Parent view object.

PR #496, opened 2018-05-31
Feature/random jira issues

Sivateja-CMI:
Copied my changes from my branches on JIRA issues.
APPDEV-1044 — Limited data fetched at the app start up. App will now fetch only data for days and weeks as directed by server url’s.
APPDEV-1075 — App is now listening to reboot and starts the background service to listen and post the app activities.
APPDEV-1069 — Environment switch options re-enables and made dynamic to take the user entered url.
APPDEV-1079 — App now uses to get the next and current data as per urls in the Links from response.

### DIFF
--- a/app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
+++ b/app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
@@ -29,4 +29,20 @@ public interface ActivityCategoryManager {
      * @return the list of activity categories
      */
     ActivityCategories getListOfActivityCategories();
+
+
+    /**
+     * Validates the new environment by requesting the activityCategoryList api. Hence the function is added in this class.
+     *  @param url the new Environment URL entered by user
+     * @param listener the listener
+     * @return the list of activity categories
+     */
+     void validateNewEnvironment(String url,final DataLoadListener listener);
+
+    /**
+     *
+     *
+     *
+     */
+     void updateNetworkAPIHost();
 }
--- a/app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
+++ b/app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
@@ -40,9 +40,8 @@ public interface ActivityCategoryManager {
      void validateNewEnvironment(String url,final DataLoadListener listener);
 
     /**
-     *
-     *
+     *Updates the network api host with Server url in App user defaults.
      *
      */
-     void updateNetworkAPIHost();
+     void updateNetworkAPIEnvironment();
 }
--- a/app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
+++ b/app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
@@ -38,6 +38,16 @@ public interface NotificationManager {
 
     void getMessage(final int itemsPerPage, final int pageNo, boolean isUnreadStatus, final DataLoadListener listener);
 
+
+    /**
+     * Gets message.
+     *
+         * @param urlForMessagesFetch the items per page
+     * @param isUnreadStatus       the page no
+     * @param listener     the listener
+     * @param isProcessUpdate
+     */
+    void getMessageWithUrl(final String urlForMessagesFetch, final boolean isUnreadStatus, final DataLoadListener listener, final boolean isProcessUpdate);
     /**
      * Post message.
      *
--- a/app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
@@ -120,4 +120,29 @@ public class ActivityCategoryManagerImpl implements ActivityCategoryManager {
         }
     }
 
+    public void updateNetworkAPIHost(){
+        activityCategoriesNetwork.updateNetworkHost();;
+    }
+
+    public void validateNewEnvironment(String url,final DataLoadListener listener) {
+        try {
+            activityCategoriesNetwork.getActivityCategories(url, new DataLoadListener() {
+                @Override
+                public void onDataLoad(Object result) {
+                    if (listener != null) {
+                        listener.onDataLoad(result);
+                    }
+                }
+
+                @Override
+                public void onError(Object errorMessage) {
+                    if (listener != null) {
+                        listener.onError(errorMessage);
+                    }
+                }
+            });
+        } catch (Exception e) {
+            AppUtils.throwException(ActivityCategoryManagerImpl.class.getSimpleName(), e, Thread.currentThread(), listener);
+        }
+    }
 }
--- a/app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
@@ -120,8 +120,8 @@ public class ActivityCategoryManagerImpl implements ActivityCategoryManager {
         }
     }
 
-    public void updateNetworkAPIHost(){
-        activityCategoriesNetwork.updateNetworkHost();;
+    public void updateNetworkAPIEnvironment(){
+        activityCategoriesNetwork.updateNeworkEnvironment();;
     }
 
     public void validateNewEnvironment(String url,final DataLoadListener listener) {
--- a/app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
@@ -27,6 +27,7 @@ import nu.yona.app.api.manager.NotificationManager;
 import nu.yona.app.api.manager.network.NotificationNetworkImpl;
 import nu.yona.app.api.model.Embedded;
 import nu.yona.app.api.model.ErrorMessage;
+import nu.yona.app.api.model.Message;
 import nu.yona.app.api.model.MessageBody;
 import nu.yona.app.api.model.Properties;
 import nu.yona.app.api.model.RegisterUser;
@@ -160,6 +161,87 @@ public class NotificationManagerImpl implements NotificationManager {
             AppUtils.throwException(NotificationManagerImpl.class.getSimpleName(), e, Thread.currentThread(), listener);
         }
     }
+
+
+    public void getMessageWithUrl(final String urlForMessagesFetch, final boolean isUnreadStatus, final DataLoadListener listener, final boolean isProcessUpdate) {
+        try {
+            User user = YonaApplication.getEventChangeManager().getDataState().getUser();
+            if (user != null && user.getLinks() != null && user.getLinks().getYonaMessages() != null
+                    && !TextUtils.isEmpty(user.getLinks().getYonaMessages().getHref())) {
+
+                notificationNetwork.getNextSetOfMessagesFromURL(urlForMessagesFetch, YonaApplication.getEventChangeManager().getSharedPreference().getYonaPassword(), isUnreadStatus, new DataLoadListener() {
+                    @Override
+                    public void onDataLoad(Object result) {
+                        if (listener != null) {
+                            if (result instanceof YonaMessages) {
+                                YonaMessages resultYonaMessagesObj = (YonaMessages) result;
+                                if (resultYonaMessagesObj != null && resultYonaMessagesObj.getEmbedded() != null) {
+                                    Embedded embedded = resultYonaMessagesObj.getEmbedded();
+                                    List<YonaMessage> listMessages = embedded.getYonaMessages();
+                                    boolean isAnyProcessed = false;
+                                    SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
+                                    for (YonaMessage message : listMessages) {
+                                        //update enum
+                                        if (message.getStatus() != null) {
+                                            message.setNotificationMessageEnum(NotificationMessageEnum.getNotificationMessageEnum(message.getType(), message.getStatus()));
+                                        } else if (message.getDropBuddyReason() != null) {
+                                            message.setNotificationMessageEnum(NotificationMessageEnum.getNotificationMessageEnum(message.getType(), message.getDropBuddyReason()));
+                                        } else {
+                                            message.setNotificationMessageEnum(NotificationMessageEnum.getNotificationMessageEnum(message.getType(), message.getChange()));
+                                        }
+                                        if (message.getLinks() != null && message.getLinks().getYonaUser() != null
+                                                && !TextUtils.isEmpty(message.getLinks().getYonaUser().getHref())) {
+                                            if (message.getEmbedded() == null) {
+                                                message.setEmbedded(new Embedded());
+                                            }
+                                            message.getEmbedded().setYonaUser(getYonaUser(message.getLinks().getYonaUser().getHref()));
+                                        }
+                                        String uploadDate = "";
+
+                                        String createdTime = message.getCreationTime();
+                                        try {
+                                            Date date = sdf.parse(createdTime);
+
+                                            Calendar futureCalendar = Calendar.getInstance();
+                                            futureCalendar.setTime(date);
+
+                                            uploadDate = DateUtility.getRelativeDate(futureCalendar);
+                                            message.setStickyTitle(uploadDate);
+                                        } catch (Exception e) {
+                                            AppUtils.throwException(ActivityManagerImpl.class.getSimpleName(), e, Thread.currentThread(), null);
+                                        }
+                                        resultYonaMessagesObj.setEmbedded(embedded);
+                                        if (!isProcessUpdate && message != null && message.getLinks() != null && message.getLinks().getYonaPreocess() != null
+                                                && !TextUtils.isEmpty(message.getLinks().getYonaPreocess().getHref())) {
+                                            isAnyProcessed = true;
+                                            MessageBody body = new MessageBody();
+                                            body.setProperties(new Properties());
+                                            postMessageForProcess(message.getLinks().getYonaPreocess().getHref(), body);
+                                        }
+                                    }
+                                    if (isAnyProcessed) {
+                                        getMessageWithUrl(urlForMessagesFetch, isUnreadStatus, listener, true);
+                                    }
+                                    APIManager.getInstance().getAuthenticateManager().getUserFromServer();
+                                }
+                                listener.onDataLoad(resultYonaMessagesObj);
+                            }
+                        }
+                    }
+
+                    @Override
+                    public void onError(Object errorMessage) {
+                        throwError(listener, errorMessage);
+                    }
+                });
+            } else {
+                listener.onError(new ErrorMessage(mContext.getString(R.string.urlnotfound)));
+            }
+        } catch (Exception e) {
+            AppUtils.throwException(NotificationManagerImpl.class.getSimpleName(), e, Thread.currentThread(), listener);
+        }
+    }
+
 
     private RegisterUser getYonaUser(String href) {
         User user = YonaApplication.getEventChangeManager().getDataState().getUser();
--- a/app/src/main/java/nu/yona/app/api/manager/network/ActivityCategoriesNetworkImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/ActivityCategoriesNetworkImpl.java
@@ -48,7 +48,7 @@ public class ActivityCategoriesNetworkImpl extends BaseImpl {
 
     }
 
-    public void updateNetworkHost(){
+    public void updateNeworkEnvironment(){
         reinitializeAPI();
     }
 }
--- a/app/src/main/java/nu/yona/app/api/manager/network/ActivityCategoriesNetworkImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/ActivityCategoriesNetworkImpl.java
@@ -47,4 +47,8 @@ public class ActivityCategoriesNetworkImpl extends BaseImpl {
         });
 
     }
+
+    public void updateNetworkHost(){
+        reinitializeAPI();
+    }
 }
--- a/app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
@@ -328,6 +328,21 @@ public interface RestApi {
     @DELETE
     Call<Void> deleteBuddy(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
 
+
+
+    /********
+     * NOTIFICATION MANAGER
+     *
+     * @param nextUrl      the url
+     * @param password the password
+     * @return the messages
+     */
+    @GET
+    Call<YonaMessages> getNextSetOfMessages(@Url String nextUrl, @Header(NetworkConstant.YONA_PASSWORD) String password,
+                                   @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage,
+                                   @Query("onlyUnreadMessages") boolean isUnreadMessage
+                                            );
+
     /********
      * NOTIFICATION MANAGER
      *
@@ -342,6 +357,20 @@ public interface RestApi {
                                    @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage,
                                    @Query("onlyUnreadMessages") boolean isUnreadMessage,
                                    @Query("size") int size, @Query("page") int page);
+
+    /********
+     * NOTIFICATION MANAGER
+     *
+     * @param url      the url
+     * @param password the password
+
+     * @return the messages
+     */
+    @GET
+    Call<YonaMessages> getAllMessages(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password,
+                                   @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage,
+                                   @Query("onlyUnreadMessages") boolean isUnreadMessage
+                                   );
 
     /**
      * Delete message call.
@@ -364,6 +393,17 @@ public interface RestApi {
      */
     @POST
     Call<Void> postMessage(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage, @Body MessageBody body);
+
+    /********
+     * ACTIVITY  @param url the url
+     *
+     * @param nextActivityUrl      the url
+     * @param password the password
+     * @return the daily activity
+     */
+    @GET
+    Call<EmbeddedYonaActivity> getNextActivityWithUrl(@Url String nextActivityUrl, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
+
 
     /********
      * ACTIVITY  @param url the url
@@ -413,6 +453,17 @@ public interface RestApi {
      *
      * @param url      the url
      * @param password the password
+     * @return the with buddy activity
+     */
+    @GET
+    Call<EmbeddedYonaActivity> getAllWithBuddyActivity(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
+
+
+    /**
+     * Gets with buddy activity.
+     *
+     * @param url      the url
+     * @param password the password
      * @param size     the size
      * @param page     the page
      * @return the with buddy activity
@@ -422,6 +473,10 @@ public interface RestApi {
 
     @GET
     Call<EmbeddedYonaActivity> getComments(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage, @Query("size") int size, @Query("page") int page);
+
+    @GET
+    Call<EmbeddedYonaActivity> getNextSetOfComments(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
+
 
     @POST
     Call<YonaMessage> addComment(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage, @Body Message message);
--- a/app/src/main/java/nu/yona/app/state/DataState.java
+++ b/app/src/main/java/nu/yona/app/state/DataState.java
@@ -22,6 +22,8 @@ import nu.yona.app.ui.BaseFragment;
 import nu.yona.app.ui.dashboard.DayActivityDetailFragment;
 import nu.yona.app.utils.AppConstant;
 
+import static nu.yona.app.YonaApplication.getEventChangeManager;
+
 /**
  * Created by kinnarvasa on 10/06/16.
  */
@@ -138,10 +140,11 @@ public class DataState {
      * @return the server url
      */
     public String getServerUrl() {
-        if (TextUtils.isEmpty(YonaApplication.getEventChangeManager().getSharedPreference().getUserPreferences().getString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.blank)))) {
-            YonaApplication.getEventChangeManager().getSharedPreference().getUserPreferences().edit().putString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.server_url)).commit();
+        if (TextUtils.isEmpty(getEventChangeManager().getSharedPreference().getAppPreferences().getString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.blank)))) {
+            getEventChangeManager().getSharedPreference().getUserPreferences().edit().putString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.server_url)).commit();
+            getEventChangeManager().getSharedPreference().getAppPreferences().edit().putString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.server_url)).commit();
         }
-        return YonaApplication.getEventChangeManager().getSharedPreference().getUserPreferences().getString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.server_url));
+        return YonaApplication.getEventChangeManager().getSharedPreference().getAppPreferences().getString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.server_url));
     }
 
     /**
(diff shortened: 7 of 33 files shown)


app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
```diff
@@ -29,4 +29,19 @@
...
     ActivityCategories getListOfActivityCategories();
+
+
+    /**
+     * Validates the new environment by requesting the activityCategoryList api. Hence the function is added in this class.
+     *  @param url the new Environment URL entered by user
+     * @param listener the listener
+     * @return the list of activity categories
+     */
+     void validateNewEnvironment(String url,final DataLoadListener listener);
```
Bert-R: Don't use ``final``, except for member variables.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
```diff
@@ -120,4 +120,29 @@ public void onError(Object errorMessage) {
         }
     }
 
+    public void updateNetworkAPIEnvironment(){
+        activityCategoriesNetwork.updateNeworkEnvironment();;
+    }
+
+    public void validateNewEnvironment(String url,final DataLoadListener listener) {
+        try {
+            activityCategoriesNetwork.getActivityCategories(url, new DataLoadListener() {
```
Bert-R: Why are you wrapping the listener you got?

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
 
+
+    public void getMessageWithUrl(final String urlForMessagesFetch, final boolean isUnreadStatus, final DataLoadListener listener, final boolean isProcessUpdate) {
+        try {
+            User user = YonaApplication.getEventChangeManager().getDataState().getUser();
+            if (user != null && user.getLinks() != null && user.getLinks().getYonaMessages() != null
+                    && !TextUtils.isEmpty(user.getLinks().getYonaMessages().getHref())) {
+
+                notificationNetwork.getNextSetOfMessagesFromURL(urlForMessagesFetch, YonaApplication.getEventChangeManager().getSharedPreference().getYonaPassword(), isUnreadStatus, new DataLoadListener() {
+                    @Override
```
Bert-R: Pull this anonymous class out of here and make it an inner class, as it makes this method way too long

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                    public void onDataLoad(Object result) {
+                        if (listener != null) {
```
Bert-R: Can this ``listener`` realistically be null?

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                            if (result instanceof YonaMessages) {
```
Bert-R: Can you realistically be called with other results than Yona messages?

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                                YonaMessages resultYonaMessagesObj = (YonaMessages) result;
+                                if (resultYonaMessagesObj != null && resultYonaMessagesObj.getEmbedded() != null) {
```
Bert-R: Can you realistically be called with a null result?

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                                    Embedded embedded = resultYonaMessagesObj.getEmbedded();
+                                    List<YonaMessage> listMessages = embedded.getYonaMessages();
+                                    boolean isAnyProcessed = false;
+                                    SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
+                                    for (YonaMessage message : listMessages) {
+                                        //update enum
+                                        if (message.getStatus() != null) {
+                                            message.setNotificationMessageEnum(NotificationMessageEnum.getNotificationMessageEnum(message.getType(), message.getStatus()));
```
Bert-R: The call to ``setNotificationMessageEnum`` is repeated three times, with different ``NotificationMessageEnum`` values. Just determine the ``NotificationMessageEnum`` value in one of the three possible ways and call ``setNotificationMessageEnum`` only once.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                    public void onDataLoad(Object result) {
+                        if (listener != null) {
+                            if (result instanceof YonaMessages) {
+                                YonaMessages resultYonaMessagesObj = (YonaMessages) result;
+                                if (resultYonaMessagesObj != null && resultYonaMessagesObj.getEmbedded() != null) {
+                                    Embedded embedded = resultYonaMessagesObj.getEmbedded();
+                                    List<YonaMessage> listMessages = embedded.getYonaMessages();
+                                    boolean isAnyProcessed = false;
+                                    SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
+                                    for (YonaMessage message : listMessages) {
```
Bert-R: Move the body of this for-loop into a separate method, to prevent too long methods.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                                        String createdTime = message.getCreationTime();
+                                        try {
+                                            Date date = sdf.parse(createdTime);
+
+                                            Calendar futureCalendar = Calendar.getInstance();
+                                            futureCalendar.setTime(date);
+
+                                            uploadDate = DateUtility.getRelativeDate(futureCalendar);
+                                            message.setStickyTitle(uploadDate);
+                                        } catch (Exception e) {
```
Bert-R: Don't catch ``Exception`` but be specific always.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                    @Override
+                    public void onDataLoad(Object result) {
+                        if (listener != null) {
+                            if (result instanceof YonaMessages) {
+                                YonaMessages resultYonaMessagesObj = (YonaMessages) result;
+                                if (resultYonaMessagesObj != null && resultYonaMessagesObj.getEmbedded() != null) {
+                                    Embedded embedded = resultYonaMessagesObj.getEmbedded();
+                                    List<YonaMessage> listMessages = embedded.getYonaMessages();
+                                    boolean isAnyProcessed = false;
+                                    SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
```
Bert-R: This is declared 20 lines before its first usage. Move it down to where you need it.
The likely reason you did this is that you don't want to instantiate the ``SimpleDateFormat`` in the loop. To prevent that, switch from ``java.util.Date`` to the date/time classes in ``java.time`` and its subpackages.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                                                && !TextUtils.isEmpty(message.getLinks().getYonaUser().getHref())) {
+                                            if (message.getEmbedded() == null) {
+                                                message.setEmbedded(new Embedded());
+                                            }
+                                            message.getEmbedded().setYonaUser(getYonaUser(message.getLinks().getYonaUser().getHref()));
+                                        }
+                                        String uploadDate = "";
+
+                                        String createdTime = message.getCreationTime();
+                                        try {
```
Bert-R: [Our coding conventions](https://wiki.yona.nu/display/DEV/Code) (which are new to you) demand that a try/catch always spans the entire method body (that's one more reason to keep the method short (<= 20 lines).

app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
```diff
@@ -38,6 +38,16 @@
...
     void getMessage(final int itemsPerPage, final int pageNo, boolean isUnreadStatus, final DataLoadListener listener);
 
+
+    /**
+     * Gets message.
+     *
+         * @param urlForMessagesFetch the items per page
+     * @param isUnreadStatus       the page no
+     * @param listener     the listener
+     * @param isProcessUpdate
```
Bert-R: What does ``isProcessUpdate`` mean?

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+                                            isAnyProcessed = true;
+                                            MessageBody body = new MessageBody();
+                                            body.setProperties(new Properties());
+                                            postMessageForProcess(message.getLinks().getYonaPreocess().getHref(), body);
+                                        }
+                                    }
+                                    if (isAnyProcessed) {
+                                        getMessageWithUrl(urlForMessagesFetch, isUnreadStatus, listener, true);
+                                    }
+                                    APIManager.getInstance().getAuthenticateManager().getUserFromServer();
```
Bert-R: Given that you're not using the return value, why are you fetching the user from the server here?

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -161,6 +162,87 @@ public void onError(Object errorMessage) {
...
+
+                    @Override
+                    public void onError(Object errorMessage) {
+                        throwError(listener, errorMessage);
+                    }
+                });
+            } else {
+                listener.onError(new ErrorMessage(mContext.getString(R.string.urlnotfound)));
+            }
+        } catch (Exception e) {
```
Bert-R: See earlier comment:
> Don't catch ``Exception`` but be specific always.

app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
Sivateja-CMI: final is used the inner classes inside the function. If I don't use final, compiler throws error to do so as there is a chance it gets modified.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Sivateja-CMI: It is a mistake. I will revisit the code part.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: I will update code here.

app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
Sivateja-CMI: This is one of the questions I had for mobiquity. I even don't see that code getting called. I didnt disturb the code as every thing is working fine.
if this is true, they are calling a post services with getYonaPreocess url with empty message body and properties.
In other places they are using same post call to set notification message as markRead.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: Will break into functions.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: Working on it.

app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
Bert-R: In this particular case, omitting final won't lead to any compiler errors, as this is in an interface. If you use anonymous inner classes, then indeed you might need to make variables final. However, you should prevent such classes as they nearly always make your method explode beyond readability constraints.

app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
Bert-R: I see. Messages in Yona can have actions, e.g. "mark read" or "accept" (on a buddy connect request). To trigger such an action, a ``POST`` needs to be done.

In the past, the app needed to do the do the action "process" on certain messages. Later, we have moved this functionality to the server. They have disabled the mechanism but not cleaned up the code. Please clean it up, as we will never revert to the old way of working.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
```diff
@@ -52,33 +53,13 @@ public ActivityCategoryManagerImpl(Context context) {
...
-            } else {
-                if (listener != null) {
-                    listener.onError(new ErrorMessage(mContext.getString(R.string.urlnotfound)));
-                }
-            }
-        } catch (Exception e) {
-            AppUtils.throwException(ActivityCategoryManagerImpl.class.getSimpleName(), e, Thread.currentThread(), listener);
+    public void getActivityCategoriesById(DataLoadListener listener) {
+        User user = YonaApplication.getEventChangeManager().getDataState().getUser();
+        if (user != null && user.getLinks() != null && user.getLinks().getYonaAppActivity() != null && !TextUtils.isEmpty(user.getLinks().getYonaAppActivity().getHref())) {
```
Bert-R: This is a terrible statement too. Can ``user` reasonably be ``null`` here? Is it reasonable to expect that the user doesn't have an app activity URL?

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
```diff
@@ -52,33 +53,13 @@ public ActivityCategoryManagerImpl(Context context) {
...
+            DataLoadListenerImpl listenerWrapper = new DataLoadListenerImpl((result) -> updateActivityCategories(result, null), listener);
+            activityCategoriesNetwork.getActivityCategories(authenticateDao.getUser().getLinks().getYonaAppActivity().getHref(), listenerWrapper);
```
Bert-R: What happens here? Why are you calling ``getActivityCategories`` with the app activity reporting URL? These seem unrelated to me.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
```diff
@@ -52,33 +53,13 @@ public ActivityCategoryManagerImpl(Context context) {
...
+        } else if (listener != null) {
+            listener.onError(new ErrorMessage(mContext.getString(R.string.urlnotfound)));
```
Bert-R: For this one, I would make a helper method on ``DataLoadListenerImpl``, so you can change this line and the above one into something like ``DataLoadListenerImpl.reportError(listener, R.string.urlnotfound);``

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: Yes. it can be null. For eg: Please see line no:226 in LoginActivity class, where we dont listen for the call back         APIManager.getInstance().getActivityCategoryManager().getActivityCategoriesById(null);

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: I don't know why they had this check. 
But I faced some issues with ErrorMessage class. Some times (When ) the call back comes with all properties to cast the object to ErrorMessage and some times it will only come with string message="some message" and will not have error code. So it better to have the check.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: i will update this in my next commit.

app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
Sivateja-CMI: I will update this in my next commit.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: i will update this in next commit

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Bert-R: OK, clear.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Sivateja-CMI: Dont know if the guys had this checks after getting some issue or they were over cautious.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Bert-R: OK

app/src/main/java/nu/yona/app/api/manager/NotificationManager.java
Bert-R: OK

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Bert-R: OK

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Bert-R: OK

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Bert-R: OK

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Bert-R: We shouldn't do that, as it's a lot of clutter.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: I will update the code here.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Sivateja-CMI: it is for sending back the response to the View class.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: Nope. I will update and remove the check and test it.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Bert-R: But it doesn't do anything except for the null-checks, right? If you put the null-check inside ``getActivityCategories``, you don't need to do it here.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: I will remove all unnecessary nested try/catch's

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Sivateja-CMI: Will debug this and see what is happening here. But it works even though they are not related!

Bert-R: Great, as this is concerning me quite a bit.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Sivateja-CMI: It passes on the result with listener.onDataLoad(result) and listener.onError(errorMessage). 
or did I not get your question properly>

Bert-R: What would go wrong if you would only do this?
```
try {
	activityCategoriesNetwork.getActivityCategories(url, listener);
} catch (Exception e) {
	AppUtils.throwException(ActivityCategoryManagerImpl.class.getSimpleName(), e, Thread.currentThread(), listener);
}
```

app/src/main/java/nu/yona/app/api/manager/ActivityCategoryManager.java
Sivateja-CMI: Lambda from your commit is implemented to get rid of final declarations.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
Sivateja-CMI: java.time package requires minimum Android API level of 19.

Bert-R: Good to hear. That's our minimum, right?

app/src/main/java/nu/yona/app/api/manager/impl/ActivityCategoryManagerImpl.java
Sivateja-CMI: URL specified in the method is ignored because of the @GET annotation. This is appending the path 'activityCategories/' to the base url of the network class. So it is working without any issues. I will remove the url argument
'''
@GET(ApiList.ACTIVITY_CATEGORIES)
    Call<ActivityCategories> getActivityCategories(@Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
'''

Bert-R: Right! That explains why it didn't fail.

app/src/main/java/nu/yona/app/api/manager/network/ActivityCategoriesNetworkImpl.java
```diff
@@ -26,13 +26,13 @@
...
      *
-     * @param url      the url
      * @param listener the listener
      */
-    public void getActivityCategories(String url, final DataLoadListener listener) {
+    public void getActivityCategories( DataLoadListener listener) {
         getRestApi().getActivityCategories(Locale.getDefault().toString().replace('_', '-')).enqueue(new Callback<ActivityCategories>() {
             @Override
             public void onResponse(Call<ActivityCategories> call, Response<ActivityCategories> response) {
+                if(listener ==null)return;
```
Bert-R: Missing curly braces

app/src/main/java/nu/yona/app/api/manager/network/ActivityCategoriesNetworkImpl.java
```diff
@@ -42,6 +42,7 @@ public void onResponse(Call<ActivityCategories> call, Response<ActivityCategorie
 
             @Override
             public void onFailure(Call<ActivityCategories> call, Throwable t) {
+                if(listener !=null);
```
Bert-R: This is a [nop statement](https://en.wikipedia.org/wiki/NOP).

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -162,82 +165,40 @@ public void onError(Object errorMessage) {
         }
     }
 
+    public YonaMessages processYonaMessages(YonaMessages result){
```
Bert-R: This improved a lot (from ~80 lines to ~30 lines), but it's still too long.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -162,82 +165,40 @@ public void onError(Object errorMessage) {
...
+        YonaMessages resultYonaMessagesObj = (YonaMessages) result;
+        Embedded embedded = resultYonaMessagesObj.getEmbedded();
+        List<YonaMessage> listMessages = embedded.getYonaMessages();
+        SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
+        for (YonaMessage message : listMessages) {
+            String messageEnumType = (message.getStatus() != null)?message.getStatus():(message.getDropBuddyReason()!=null)?message.getDropBuddyReason():message.getChange();
```
Bert-R: Move the body of this for-loop into a separate method: ``processYonaMessage``.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -162,82 +165,40 @@ public void onError(Object errorMessage) {
...
+            message.setNotificationMessageEnum(NotificationMessageEnum.getNotificationMessageEnum(message.getType(), messageEnumType));
+            if (message.getLinks() != null && message.getLinks().getYonaUser() != null && !TextUtils.isEmpty(message.getLinks().getYonaUser().getHref())) {
+                if (message.getEmbedded() == null) {
+                    message.setEmbedded(new Embedded());
+                }
+                message.getEmbedded().setYonaUser(getYonaUser(message.getLinks().getYonaUser().getHref()));
+            }
+            String uploadDate = "";
+            String createdTime = message.getCreationTime();
```
Bert-R: Move this line and the one above inside the try/catch-block and then extract the entire try/catch-block into a separate method. Move the creation of the ``SimpleDateFormat`` into that method too.

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -162,82 +165,40 @@ public void onError(Object errorMessage) {
         }
     }
 
+    public YonaMessages processYonaMessages(YonaMessages result){
+        YonaMessages resultYonaMessagesObj = (YonaMessages) result;
```
Bert-R: Remove ``Obj`` from this variable name

app/src/main/java/nu/yona/app/api/manager/impl/NotificationManagerImpl.java
```diff
@@ -162,82 +165,40 @@ public void onError(Object errorMessage) {
...
+        Embedded embedded = resultYonaMessagesObj.getEmbedded();
+        List<YonaMessage> listMessages = embedded.getYonaMessages();
+        SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
+        for (YonaMessage message : listMessages) {
```
Bert-R: Inline the lines 170 and 171 here:
```
for (YonaMessage message : resultYonaMessagesObj.getEmbedded().getYonaMessages()) {
```

app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
```diff
@@ -328,6 +328,21 @@
...
+
+    /********
+     * NOTIFICATION MANAGER
+     *
+     * @param nextUrl      the url
+     * @param password the password
+     * @return the messages
+     */
+    @GET
+    Call<YonaMessages> getNextSetOfMessages(@Url String nextUrl, @Header(NetworkConstant.YONA_PASSWORD) String password,
```
Bert-R: This simply gets the messages from the URL, whether the first page or the next isn't known here, so rename into ``getMessages``.

app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
```diff
@@ -365,6 +394,17 @@
...
 
+    /********
+     * ACTIVITY  @param url the url
+     *
+     * @param nextActivityUrl      the url
+     * @param password the password
+     * @return the daily activity
+     */
+    @GET
+    Call<EmbeddedYonaActivity> getNextActivityWithUrl(@Url String nextActivityUrl, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
```
Bert-R: This fetches a page of activities. Whether the first or the next is unknown, so just call it ``getActivities``.

app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
```diff
@@ -408,6 +448,17 @@
...
 
+    /**
+     * Gets with buddy activity.
+     *
+     * @param url      the url
+     * @param password the password
+     * @return the with buddy activity
+     */
+    @GET
+    Call<EmbeddedYonaActivity> getAllWithBuddyActivity(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
```
Bert-R: This isn't being used.

app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
```diff
@@ -421,7 +472,11 @@
     Call<EmbeddedYonaActivity> getWithBuddyActivity(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage, @Query("size") int size, @Query("page") int page);
 
     @GET
-    Call<EmbeddedYonaActivity> getComments(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage, @Query("size") int size, @Query("page") int page);
+    Call<EmbeddedYonaActivity> getComments(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanaguage);
+//
```
Bert-R: Remove commented code.

app/src/main/java/nu/yona/app/state/DataState.java
```diff
@@ -138,10 +140,11 @@ public void setEmbeddedWithBuddyActivity(EmbeddedYonaActivity embeddedWithBuddyA
      * @return the server url
      */
     public String getServerUrl() {
-        if (TextUtils.isEmpty(YonaApplication.getEventChangeManager().getSharedPreference().getUserPreferences().getString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.blank)))) {
-            YonaApplication.getEventChangeManager().getSharedPreference().getUserPreferences().edit().putString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.server_url)).commit();
+        if (TextUtils.isEmpty(getEventChangeManager().getSharedPreference().getAppPreferences().getString(AppConstant.SERVER_URL, YonaApplication.getAppContext().getString(R.string.blank)))) {
```
Bert-R: Deduplicate this method and the next one.

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -223,28 +247,36 @@ public void onDestroyView() {
...
     public void onResume() {
         super.onResume();
-        if (activity != null) {
-            setDayActivityDetails();
-        }
-        if (viewPager != null) {
-            viewPager.addOnPageChangeListener(new ViewPager.OnPageChangeListener() {
-                @Override
-                public void onPageScrolled(int position, float positionOffset, int positionOffsetPixels) {
+//        if (activity != null) {
```
Bert-R: Remove commented code

app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
```diff
@@ -243,6 +252,9 @@ public void onPageScrollStateChanged(int state) {
         return view;
     }
 
+
```
Bert-R: Remove empty lines

app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
```diff
@@ -261,32 +273,59 @@ public void onDestroyView() {
...
+                } else {
+                    goBack();
+                }
             } else {
                 goBack();
             }
-        } else {
-            goBack();
+            setDayDetailTitleAndIcon();
+        } catch (NullPointerException e) {
```
Bert-R: Catching an NPE is a code smell. Why is it necessary here?

app/src/main/java/nu/yona/app/ui/signup/SignupActivity.java
```diff
@@ -225,15 +225,19 @@ private void showMobileVerificationScreen(Bundle bundle) {
...
-        if (message.getCode() != null && (message.getCode().equalsIgnoreCase(ServerErrorCode.USER_EXIST_ERROR)
-                || message.getCode().equalsIgnoreCase(ServerErrorCode.ADD_BUDDY_USER_EXIST_ERROR))) {
-            showAlertForReRegisteruser(message.getMessage());
-        } /*else if(message.getCode() != null && (message.getCode().equalsIgnoreCase(ServerErrorCode.SMS_SENDING_FAIL))) {
+        if (errorMessage instanceof ErrorMessage){
+            ErrorMessage message = (ErrorMessage) errorMessage;
+            if (message.getCode() != null && (message.getCode().equalsIgnoreCase(ServerErrorCode.USER_EXIST_ERROR)
+                    || message.getCode().equalsIgnoreCase(ServerErrorCode.ADD_BUDDY_USER_EXIST_ERROR))) {
+                showAlertForReRegisteruser(message.getMessage());
+            } /*else if(message.getCode() != null && (message.getCode().equalsIgnoreCase(ServerErrorCode.SMS_SENDING_FAIL))) {
```
Bert-R: Remove commented code

app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
Sivateja-CMI: We have removed quite a few null checks in the inner classes as part of re-factoring. So I thought of adding it to avoid the crashes.

Sivateja-CMI: While correcting code with your earlier reviews, I have removed (embeddedYonaActivity != null) check which caused the crash on launch. So added it to avoid the crash. We can remove it now as code is safe with null check.

PR #497, opened 2018-06-23
APPDEV-1137 Correct support email address

Bert-R:
Currently, the address is said to be app@yona.nu, but it should be support@yona.nu


Bert-R:
If you're OK with it, please go ahead and merge it. Please always use the option "Squash and merge" option of GitHub, so you're adding all commits as a single one to the ``develop`` branch. In this particular case, it doesn't really matter as it's just one commit, but just make it a habit.

PR #499, opened 2018-08-06
APPDEV-1142 Implement versioning strategy

Bert-R:
See [APPDEV-1142](https://jira.yona.nu/browse/APPDEV-1142) for what's to be accomplished.


app/build.gradle
```diff
@@ -1,7 +1,7 @@
 apply plugin: 'com.android.application'
 
 android {
-    def versionCode = 0
+    def newVersionCode = 0
 	if (System.env.BRANCH_NAME) {
```
Sivateja-CMI: I feel we can remove this check as file is now having same name always.

Bert-R: If we would remove this condition, then local builds (outside Jenkins) would update the version number, which would be wrong.

Sivateja-CMI: Ok. got it.

PR #504, opened 2018-08-16
APPDEV-1152 Implement fastlane deployment

Bert-R:
The deployment steps are actually empty, as we don't have the keystore file yet to do the upload. We'll fill that in once it's done.

Along with the actual changes, we removed YONA-Automation, as nobody used it and we reduced the build flavors to just two: develop and production.

Just an interesting note: the build now runs in a Docker container on our Jenkins agent. That keeps the agent clean and standard.


Sivateja-CMI:
Merged the PR.  Thanks @Bert-R. Great to have a complete pipeline for Android.

PR #507, opened 2018-08-23
APPDEV-1139 inform server when app is open

Sivateja-CMI:
- Implemented Functionality to post yona open app event to server.
- App posts the event when user opens the app.

### DIFF
--- a/app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
@@ -24,6 +24,7 @@ import nu.yona.app.api.model.ErrorMessage;
 import nu.yona.app.api.utils.NetworkUtils;
 import nu.yona.app.api.utils.ServerErrorCode;
 import nu.yona.app.listener.DataLoadListener;
+import nu.yona.app.listener.DataLoadListenerImpl;
 import nu.yona.app.state.EventChangeManager;
 import nu.yona.app.utils.AppUtils;
 import okhttp3.Interceptor;
@@ -138,6 +139,30 @@ private final Interceptor getInterceptor = new Interceptor() {
      * @return the call
      */
     Callback getCall(final DataLoadListener listener) {
+        return new Callback() {
+            @Override
+            public void onResponse(retrofit2.Call call, retrofit2.Response response) {
+                if (response.code() < NetworkConstant.RESPONSE_STATUS && listener != null) {
+                    listener.onDataLoad(response.body());
+                } else {
+                    onError(response, listener);
+                }
+            }
+
+            @Override
+            public void onFailure(retrofit2.Call call, Throwable t) {
+                onError(t, listener);
+            }
+        };
+    }
+
+    /**
+     * Gets call.
+     *
+     * @param listener the listener
+     * @return the call
+     */
+    Callback getCall(final DataLoadListenerImpl listener) {
         return new Callback() {
             @Override
             public void onResponse(retrofit2.Call call, retrofit2.Response response) {
--- a/app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
@@ -13,6 +13,7 @@ package nu.yona.app.api.manager.network;
 import nu.yona.app.api.model.ActivityCategories;
 import nu.yona.app.api.model.AddBuddy;
 import nu.yona.app.api.model.AppActivity;
+import nu.yona.app.api.model.AppMetaInfo;
 import nu.yona.app.api.model.DayActivity;
 import nu.yona.app.api.model.EmbeddedYonaActivity;
 import nu.yona.app.api.model.Goals;
@@ -419,6 +420,18 @@ public interface RestApi {
      */
     @POST
     Call<Void> postAppActivity(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanguage, @Body AppActivity activity);
+
+    /********
+     * APP OPEN EVENT
+     *
+     * @param url      the url
+     * @param password the password
+     * @param appMetaInfo the activity
+     * @return the call
+     */
+    @POST
+    Call<Void> postOpenAppEvent(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanguage, @Body AppMetaInfo appMetaInfo);
+
 
     /**
      * Gets with buddy activity.
--- a/app/src/main/java/nu/yona/app/api/model/AppMetaInfo.java
+++ b/app/src/main/java/nu/yona/app/api/model/AppMetaInfo.java
@@ -1,0 +1,46 @@
+package nu.yona.app.api.model;
+
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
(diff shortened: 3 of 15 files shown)


app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
```diff
@@ -155,6 +156,30 @@ public void onFailure(retrofit2.Call call, Throwable t) {
         };
     }
 
+    /**
+     * Gets call.
+     *
+     * @param listener the listener
+     * @return the call
+     */
+    Callback getCall(final DataLoadListenerImpl listener) {
```
Bert-R: Please rename in ``createCallback``. Get suggest it already exists and you're making a callback, not a call.

app/src/main/java/nu/yona/app/api/manager/network/RestApi.java
```diff
@@ -420,6 +421,18 @@
...
+    /********
+     * APP OPEN EVENT
+     *
+     * @param url      the url
+     * @param password the password
+     * @param appMetaInfo the activity
+     * @return the call
+     */
+    @POST
+    Call<Void> postOpenAppEvent(@Url String url, @Header(NetworkConstant.YONA_PASSWORD) String password, @Header(NetworkConstant.ACCEPT_LAUNGUAGE) String acceptLanguage, @Body AppMetaInfo appMetaInfo);
```
Bert-R: Please rename ``ACCEPT_LAUNGUAGE`` into ``ACCEPT_LANGUAGE``.

app/src/main/java/nu/yona/app/api/model/AppMetaInfo.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public String getAppVersion() {
+        return appVersion;
+    }
+
+    public String getOperatingSystem() {
+        return operatingSystem;
+    }
+
```
Bert-R: I suggest a slightly different approach. The files ``SettingsFragment.java`` and ``AppNetworkImpl.java`` contains this statement:

```
AppMetaInfo appMetaInfo = new AppMetaInfo();
```

That looks odd, as nothing goes into the constructor.

To prevent that, please do the following:
* Make a private constructor that takes three parameters: ``private AppMetaInfo (String operatingSystem, String appVersion, int appVersionCode)``
* Assign the parameters to the member variables
* Create a private static member: ``private static final AppMetaInfo theInstance = new AppMetaInfo ("ANDROID", BuildConfig.VERSION_NAME, BuildConfig.VERSION_CODE);``
* Create a public static method: ``private static AppMetaInfo()`` that returns the instance
* Use that in the aforementioned two files

app/src/main/java/nu/yona/app/api/model/AppMetaInfo.java
```diff
@@ -0,0 +1,46 @@
...
+import nu.yona.app.utils.AppConstant;
+import nu.yona.app.utils.AppUtils;
+
+import static nu.yona.app.ui.YonaActivity.getActivity;
+
+public class AppMetaInfo {
+
+    @SerializedName("operatingSystem")
+    @Expose
+    private String operatingSystem = YonaApplication.getAppContext().getString(R.string.operating_system);
```
Bert-R: There is no value in using a resource for this. It's an enum value, not a translatable string.

app/src/main/java/nu/yona/app/api/model/AppMetaInfo.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @SerializedName("appVersion")
+    @Expose
+    private String appVersion  =  BuildConfig.VERSION_NAME;
+
+    @SerializedName("appVersionCode")
+    @Expose
+    private Integer appVersionCode = BuildConfig.VERSION_CODE;
```
Bert-R: Please make this ``int``

app/src/main/res/values-nl/strings.xml
```diff
@@ -219,5 +219,7 @@
     <string name="new_environment_switch_success_msg">Je gebruikt nu :</string>
     <string name="environment_alert_title">Voer de server URL in :</string>
     <string name="enviroment_input_placeholder">Enter Text Here</string>
+    <string name="generic_alert_title">Alarm</string>
+    <string name="operating_system">ANDROID</string>
```
Bert-R: This one we do not need.

app/src/main/res/values-nl/strings.xml
```diff
@@ -219,5 +219,7 @@
     <string name="new_environment_switch_success_msg">Je gebruikt nu :</string>
     <string name="environment_alert_title">Voer de server URL in :</string>
     <string name="enviroment_input_placeholder">Enter Text Here</string>
+    <string name="generic_alert_title">Alarm</string>
```
Bert-R: Make this Waarschuwing

app/src/main/res/values/strings.xml
```diff
@@ -221,4 +221,6 @@
     <string name="new_environment_switch_success_msg">You are now using :</string>
     <string name="environment_alert_title">Enter server URL :</string>
     <string name="enviroment_input_placeholder">Enter Text Here</string>
+    <string name="generic_alert_title">Alert</string>
+    <string name="operating_system">ANDROID</string>
```
Bert-R: We don't need this one.

app/src/main/java/nu/yona/app/api/model/AppMetaInfo.java
Sivateja-CMI: Do you mean to get instance like this ?
 `AppMetaInfo appMetaInfo = AppMetaInfo.AppMetaInfo();`

app/src/main/java/nu/yona/app/ui/settings/SettingsFragment.java
```diff
@@ -84,11 +85,11 @@ public void onItemClick(AdapterView<?> parent, View view, int position, long id)
                 }
             }
         });
-        try {
-            PackageInfo pInfo = getActivity().getPackageManager().getPackageInfo(getActivity().getPackageName(), 0);
-            ((TextView) view.findViewById(R.id.label_version)).setText(getString(R.string.version) + pInfo.versionName + getString(R.string.space) + pInfo.versionCode);
-        } catch (PackageManager.NameNotFoundException e) {
-            AppUtils.throwException(SettingsFragment.class.getSimpleName(), e, Thread.currentThread(), null);
+        AppMetaInfo appMetaInfo = AppMetaInfo.getInstance();
+        if(appMetaInfo.getAppVersionCode() != 0){
```
Bert-R: What is the purpose of this if-statement?

Sivateja-CMI: My intention was to avoid crash, if values are not set for reason during build time

Bert-R: I'm not in favor of this type of constructs as it complicates the code for no reason. This data comes from the Android metadata, so it's always there. If we ever manage to create a broken build, it should be flagged rather than accepted.

Can you take this construct out and just rely on the metadata?

Sivateja-CMI: Ok. I will remove it.

Bert-R: Can you take this along in your current PR?

Sivateja-CMI: Done this as part of https://github.com/yonadev/yona-app-android/pull/513

PR #515, opened 2018-09-14
APPDEV-1170 Crashlytics crash report in apputils class

Sivateja-CMI:
-- Added support for showing user Toast when exception occurs in non-main threads.

### DIFF
--- a/app/src/main/java/nu/yona/app/api/model/ErrorMessage.java
+++ b/app/src/main/java/nu/yona/app/api/model/ErrorMessage.java
@@ -20,17 +20,17 @@ public class ErrorMessage extends BaseEntity
 {
 	@SerializedName("message")
 	@Expose
-	private String message;
+	private final String message;
 	@SerializedName("code")
 	@Expose
-	private String code;
+	private final String code;
 
 	/**
 	 * Instantiates a new Error message.
 	 */
 	public ErrorMessage()
 	{
-
+		this(null, null);
 	}
 
 	/**
@@ -40,8 +40,7 @@ public class ErrorMessage extends BaseEntity
 	 */
 	public ErrorMessage(String message)
 	{
-		this.code = "";
-		this.message = message;
+		this(message, null);
 	}
 
 
--- a/app/src/main/java/nu/yona/app/api/model/ErrorMessage.java
+++ b/app/src/main/java/nu/yona/app/api/model/ErrorMessage.java
@@ -44,6 +44,20 @@ public class ErrorMessage extends BaseEntity
 		this.message = message;
 	}
 
+
+	/**
+	 * Instantiates a new Error message.
+	 *
+	 * @param message the message
+	 * @param code    the error code for custom messages.
+	 */
+	public ErrorMessage(String message, String code)
+	{
+		this.code = code;
+		this.message = message;
+	}
+
+
 	/**
 	 * Gets message.
 	 *
@@ -52,16 +66,6 @@ public class ErrorMessage extends BaseEntity
 	public String getMessage()
 	{
 		return message;
-	}
-
-	/**
-	 * Sets message.
-	 *
-	 * @param message The message
-	 */
-	public void setMessage(String message)
-	{
-		this.message = message;
 	}
 
 	/**
@@ -74,15 +78,6 @@ public class ErrorMessage extends BaseEntity
 		return code;
 	}
 
-	/**
-	 * Sets code.
-	 *
-	 * @param code The code
-	 */
-	public void setCode(String code)
-	{
-		this.code = code;
-	}
 
 	@Override
 	public ContentValues getDbContentValues()
--- a/app/src/main/java/nu/yona/app/utils/AppUtils.java
+++ b/app/src/main/java/nu/yona/app/utils/AppUtils.java
@@ -306,7 +306,7 @@ public class AppUtils
 	 * getErrorMessage from give Exception or return a generic error message
 	 */
 
-	public static ErrorMessage getErrorMessageFromException(Exception exception)
+	private static ErrorMessage getErrorMessageFromException(Exception exception)
 	{
 		if (exception != null && exception.getMessage() != null)
 		{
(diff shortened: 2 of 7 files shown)


app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -288,25 +289,37 @@ public static String getHourDigit(String digit)
...
+			logExceptionToCrashlytics(e);
 		}
-		if (listener != null)
+	}
+
+	/**
+	 * getErrorMessage from give Exception or return a generic error message
+	 */
+
+	public static ErrorMessage getErrorMessage(Exception e)
```
Bert-R: This method doesn't have to be public.

Bert-R: Generally don't use such short names. Exceptions to that rule are: lambda expressions, loop iterator value,  name of exception in catch block

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -288,25 +289,37 @@ public static String getHourDigit(String digit)
...
+	{
+		ErrorMessage errorMessage = new ErrorMessage();
+		if (e != null && e.getMessage() != null)
 		{
-			listener.onError(errorMessage);
+			errorMessage.setMessage(e.getMessage());
+			Logger.loge(e.getClass().getSimpleName(), e.getMessage());
```
Bert-R: This shouldn't be part of ``getErrorMessage``. A ``get`` method should just get something.
Move the logging to ``reportException``

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -288,25 +289,37 @@ public static String getHourDigit(String digit)
...
 		}
 		else
 		{
-			showErrorToast(errorMessage);
-			logExceptionToCrashlytics(e);
+			String message = YonaApplication.getAppContext().getString(R.string.generic_error_message);
+			errorMessage.setMessage(message);
+			Logger.loge(e.getClass().getSimpleName(), message);
```
Bert-R: This line is nearly identical to the one I just commented on. It could easily be put below the  if/else and take the message text from the just constructed error message. Now it should be removed, as said in the previous comment.

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -288,25 +289,37 @@ public static String getHourDigit(String digit)
...
 		}
-		Logger.loge(e.getClass().getSimpleName(), e.getMessage());
+		return errorMessage;
```
Bert-R: Please make the method this simple:
```
	public static ErrorMessage getErrorMessage(Exception e)
	{
		ErrorMessage errorMessage = new ErrorMessage();
		if (e != null && e.getMessage() != null)
		{
			return new ErrorMessage(e.getMessage());
		}
		return new ErrorMessage(YonaApplication.getAppContext().getString(R.string.generic_error_message));
	}
```

Please also add ``ErrorMessage`` to your refactoring list: this is a perfect class for immutable behavior. Just remove all setters and only allow taking the values into the constructor.

app/src/main/res/values-nl/strings.xml
```diff
@@ -222,5 +222,6 @@
     <string name="generic_alert_title">Waarschuwing</string>
     <string name="generic_exception_message">Onbekende fout. Neem contact op met Yona support.</string>
     <string name="generic_mobile_number_format_error">Ongeldig telefoonnummer. Controleer het formaat</string>
+    <string name="generic_error_message">Onbekende fout. Neem contact op met Yona support.</string>
```
Bert-R: This message already exists on line 223.

app/src/main/res/values/strings.xml
```diff
@@ -224,4 +224,5 @@
     <string name="generic_alert_title">Alert</string>
     <string name="generic_exception_message">Unknown Exception. Please contact Yona support</string>
     <string name="generic_mobile_number_format_error">Invalid Mobile Number. Please check the format</string>
+    <string name="generic_error_message">Unknown Error. Please contact Yona support</string>
```
Bert-R: This message already exists on line 225.

app/src/main/java/nu/yona/app/api/model/ErrorMessage.java
```diff
@@ -74,15 +78,6 @@ public String getCode()
 		return code;
 	}
 
-	/**
-	 * Sets code.
-	 *
-	 * @param code The code
-	 */
-	public void setCode(String code)
```
Bert-R: With the setters gone now, please make the members ``final``.

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -282,53 +283,58 @@ public static String getHourDigit(String digit)
...
 		}
-		Logger.loge(e.getClass().getSimpleName(), e.getMessage());
+		Logger.loge(className, errorMessage.getMessage());
+	}
+
+	/**
+	 * getErrorMessage from give Exception or return a generic error message
+	 */
+
+	public static ErrorMessage getErrorMessageFromException(Exception exception)
```
Bert-R: Make this one ``private``.

app/src/test/java/nu/yona/app/api/manager/impl/AuthenticateManagerImplTest.java
```diff
@@ -34,211 +31,240 @@
...
-                ArgumentMatchers.any(DataLoadListener.class));
-
-        Mockito.when(authenticateDAOMock.getUser()).thenReturn(getMockedUser());
-    }
-
-    @After
-    public void tearDown() {
-        dbhelper.close();
-    }
+public class AuthenticateManagerImplTest extends YonaTestCase
```
Bert-R: I see I didn't reformat the tests. Sorry about that. Can you do a separate PR with a reformat of the remaining tests?

Sivateja-CMI: ok. will do that

PR #520, opened 2018-10-04
APPDEV-1177 Enhanced texts

Bert-R:
Seeing the texts in context, I believe they need to be enhanced


Sivateja-CMI:
I am fine with the changes.

PR #522, opened 2018-10-08
APPDEV-1182 Notifications list display is not identical in both iOS and Android

Sivateja-CMI:
-- If user has a photo, it will be used else will have Text into image view.
-- Added missing messages titles for buddy request, acceptance and rejection.
As per [UX](https://wiki.yona.nu/display/DEV/Me+-+Notifications) updated android notifications list screen.


app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+			setUpMessageItemHolderDetails();
+			currentMessageItemHolder.deleteMsg.setTag(currentYonaMessage);
+			currentMessageItemHolder.messageContainer.setTag(currentYonaMessage);
+		}
+	}
 
-		if (yonaObject != null)
+	private void setUpMessageListItemTitle()
+	{
+		if (currentYonaMessage.getNotificationMessageEnum() != null && !TextUtils.isEmpty(currentYonaMessage.getNotificationMessageEnum().getUserMessage()))
```
Bert-R: It should never be possible that ``getNotificationMessageEnum`` returns null. Assign ``NONE_NONE`` as default value in ``YonaMessage``, and you'll never need to check for null again.

Bert-R: As far as I can see, ``currentYonaMessage.getNotificationMessageEnum().getUserMessage()`` never returns an empty string. Do you agree to that or am I overlooking something?

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
 		{
-			if (yonaObject.getNotificationMessageEnum() != null && !TextUtils.isEmpty(yonaObject.getNotificationMessageEnum().getUserMessage()))
-			{
-				holder.txtTitleMsg.setText(yonaObject.getMessage());
-				holder.img_status.setImageResource(yonaObject.getNotificationMessageEnum().getImageId());
-			}
-			if (yonaObject.getLinks() != null && yonaObject.getLinks().getEdit() != null)
+			if (currentYonaMessage.getNotificationMessageEnum().getUserMessage().equals(NotificationEnum.BUDDYCONNECTREQUESTMESSAGE))
```
Bert-R: This will never be true. ``currentYonaMessage.getNotificationMessageEnum().getUserMessage()`` returns a string, which will never be equal to an enum value. It should be this:
```
if (currentYonaMessage.getNotificationMessageEnum() == NotificationEnum.BUDDYCONNECTREQUESTMESSAGE)
```
(enums can be compared with ``==``).

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
-			{
-				holder.txtFooterMsg.setText(yonaObject.getNickname());
-				if (yonaObject.getNotificationMessageEnum() == NotificationMessageEnum.GOALCONFLICTMESSAGE_ANNOUNCED)
+			currentMessageItemHolder.img_status.setImageResource(currentYonaMessage.getNotificationMessageEnum().getImageId());
+		}
+	}
+
+	private String getMessageListItemTitle(NotificationEnum yonaNotificationEnum)
+	{
+		String messagetitle = "";
```
Bert-R: No need to have this variable.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+		switch (yonaNotificationEnum)
+		{
+			case BUDDYCONNECTREQUESTMESSAGE:
+				messagetitle = this.activity.getString(R.string.buddyconnectrequested);
+				break;
```
Bert-R: Just replace the above two lines with this:
```
return this.activity.getString(R.string.buddyconnectrequested);
```

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+			case BUDDYCONNECTRESPONSEMESSAGE:
+				if (currentYonaMessage.getNotificationMessageEnum().getStatusEnum() == ACCEPTED)
 				{
-					holder.img_avtar.setImageResource(R.drawable.adult_sad);
-					holder.img_avtar.setVisibility(View.VISIBLE);
-					holder.profileIconTxt.setVisibility(View.GONE);
+					messagetitle = this.activity.getString(R.string.buddyconnectresponseaccepted);
```
Bert-R: Similar here.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
 				}
-				else
+				else if (currentYonaMessage.getNotificationMessageEnum().getStatusEnum() == REJECTED)
```
Bert-R: Given the above comment, there is no need for an ``else`` here.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
 				{
-					holder.img_avtar.setVisibility(View.GONE);
-					holder.profileIconTxt.setVisibility(View.VISIBLE);
-					holder.profileIconTxt.setText(yonaObject.getNickname().substring(0, 1).toUpperCase());
-					holder.profileIconTxt.setBackground(ContextCompat.getDrawable(YonaActivity.getActivity(), R.drawable.bg_small_self_round));
+					messagetitle = this.activity.getString(R.string.buddyconnectresponserejected);
```
Bert-R: Just return from here.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
-			{
-				holder.messageContainer.setBackground(ContextCompat.getDrawable(activity, R.drawable.item_gradient));
-			}
-			holder.deleteMsg.setTag(yonaObject);
-			holder.messageContainer.setTag(yonaObject);
+				break;
+			default:
+				break;
 		}
+		return messagetitle;
```
Bert-R: Replace the lines 118 - 122 with just this:
```
	// Fall-through intended
}
throw new IllegalArgumentException("Cannot produce an item title for " + yonaNotificationEnum);
```

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+	}
+
+	private void setUpMessageItemHolderSwipe()
+	{
+		if (currentYonaMessage.getLinks() != null && currentYonaMessage.getLinks().getEdit() != null)
```
Bert-R: You can replace the entire method body with this:
```
boolean isEditable = currentYonaMessage.getLinks() != null && currentYonaMessage.getLinks().getEdit() != null;
currentMessageItemHolder.swipeLayout.setRightSwipeEnabled(isEditable);
```

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,137 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+			currentMessageItemHolder.img_avtar.setVisibility(View.GONE);
+			currentMessageItemHolder.profileIconTxt.setVisibility(View.VISIBLE);
+			currentMessageItemHolder.profileIconTxt.setText(currentYonaMessage.getNickname().substring(0, 1).toUpperCase());
+			currentMessageItemHolder.profileIconTxt.setBackground(ContextCompat.getDrawable(YonaActivity.getActivity(), R.drawable.bg_small_self_round));
+		}
+	}
+
+	private void setUpMessageContainerBackground()
+	{
+		if (currentYonaMessage.getLinks() != null && currentYonaMessage.getLinks().getMarkRead() != null && !TextUtils.isEmpty(currentYonaMessage.getLinks().getMarkRead().getHref()))
```
Bert-R: This could be made quite a bit more readable (no need to check whether ``getHref`` returns non-null):
```
boolean isUnread = currentYonaMessage.getLinks() != null && currentYonaMessage.getLinks().getMarkRead() != null;
int resourceId = (isUnread) ? R.drawable.item_selected_gradient : R.drawable.item_gradient;
currentMessageItemHolder.messageContainer.setBackground(ContextCompat.getDrawable(activity, resourceId));
```

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Bert-R: That's not what's being compared here: The message returned in the JSON does not end up in the enum. As far as I can see, the user message in the enum is always a nonempty string.

Sivateja-CMI: Sorry. I am being stupid here. I will correct it. Notification Enums are declared with user messages hard-coded.

Sivateja-CMI: Done.

Sivateja-CMI: Corrected code. No needed all those checks as ```NotificationMessageEnum``` always has ```userMessage```.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Sivateja-CMI: Code updated as the above code to set message title is not needed. 
User Message from NotificationMessageEnum alone is sufficient.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Sivateja-CMI: Done.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Sivateja-CMI: Done.

app/src/main/java/nu/yona/app/enums/NotificationMessageEnum.java
```diff
@@ -20,7 +20,7 @@
 	/**
 	 * Buddy connect request accepted notification message enum.
 	 */
-	BUDDY_CONNECT_REQUEST_ACCEPTED(NotificationEnum.BUDDYCONNECTREQUESTMESSAGE, StatusEnum.ACCEPTED, getString(R.string.buddyconnectrequested), R.drawable.icn_ok),
+	BUDDY_CONNECT_REQUEST_ACCEPTED(NotificationEnum.BUDDYCONNECTREQUESTMESSAGE, StatusEnum.ACCEPTED, getString(R.string.buddyresponseaccepted), R.drawable.icn_ok),
```
Bert-R: These resource names are very confusing. Let's look at what pairs exist:
* There are connect requests and connect response messages.
* The status can be accepted or rejected
* There are messages sent by you and sent to you

The flow is like this:
* Richard sends a connect request message to Bob. Richard now has no messages, Bob has one. The status in that message is "requested".
* Bob now can accept or reject that message. This will cause the status of that message to become accepted or rejected and Richard will receive a buddy connect response message with the status accepted or rejected.

Given this, I propose to define the following enum values resources:
* ``BUDDY_CONNECT_REQUEST_OPEN``, ``buddyrequestopen``
* ``BUDDY_CONNECT_REQUEST_ACCEPTED``, ``buddyrequestaccepted``
* ``BUDDY_CONNECT_REQUEST_REJECTED``, ``buddyrequestrejected``
* ``BUDDY_CONNECT_RESPONSE_ACCEPTED``, ``buddyresponseaccepted``
* ``BUDDY_CONNECT_RESPONSE_REJECTED``, ``buddyresponserejected``

Makes sense?

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,107 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+			currentMessageItemHolder.profileIconTxt.setText(getMessageProfileIconText());
+			currentMessageItemHolder.profileIconTxt.setBackground(ContextCompat.getDrawable(YonaActivity.getActivity(), R.drawable.bg_small_self_round));
+		}
+	}
+
+	private String getMessageProfileIconText()
+	{
+		RegisterUser registerUser = currentYonaMessage.getEmbedded().getYonaUser();
+		User loggedInUser = getEventChangeManager().getDataState().getUser();
+		if (loggedInUser.getMobileNumber() != registerUser.getMobileNumber())
```
Bert-R: I don't think this if-statement is doing the right thing. The icon should never reflect yourself. If you are the sender, the icon should be of the target. If you are the receiver, the icon should be of the sender.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
```diff
@@ -64,63 +71,107 @@ public MessageItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType)
...
+			currentMessageItemHolder.img_avtar.setVisibility(View.GONE);
+			currentMessageItemHolder.profileIconTxt.setVisibility(View.VISIBLE);
+			currentMessageItemHolder.profileIconTxt.setText(getMessageProfileIconText());
+			currentMessageItemHolder.profileIconTxt.setBackground(ContextCompat.getDrawable(YonaActivity.getActivity(), R.drawable.bg_small_self_round));
+		}
+	}
+
+	private String getMessageProfileIconText()
+	{
+		RegisterUser registerUser = currentYonaMessage.getEmbedded().getYonaUser();
```
Bert-R: This is tricky. Not all messages have an embedded user. More often than not, the message contains a link to the user, which you then need to find among the buddies of the user.

Next, the name ``registerUser`` isn't very meaningful. Please make it ``senderUser``.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Sivateja-CMI: > If you are the sender, the icon should be of the target. If you are the receiver, the icon should be of the sender.

Is it possible that I am the sender(event actor) and I will get notified of the same?

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Sivateja-CMI: what should I use when there is not ```yona:user``` present in the reponse.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Bert-R: The majority of the messages in your message list is sent to you, but when having a conversation on an activity, the list also contains messages sent by yourself to someone.

So the rule to define the icon should be: if I'm the sender, show the receiver, otherwise the sender.

app/src/main/java/nu/yona/app/ui/message/MessageStickyRecyclerAdapter.java
Bert-R: As far as I know, there is always either a ``yona:user`` link or an embedded user.

Bert-R: 👍

Bert-R: You could add an enum to your notification enum type indicating one of the following three options:
* No user (for ``SystemMessage`` and for ``NONE_NONE``)
* Embedded user (``BuddyConnectRequest`` and ``BuddyDisconnect``)
* Linked user (others)

Sivateja-CMI: Can I directly use Nickname associated with each notification object instead. As this is required field?

Bert-R: Yes, you can use that. Sorry for making it more complex than necessary.

Bert-R: Wait. The avatar is built from the first and last name, so the nickname is of no help here.

Sivateja-CMI: > Avatar section https://wiki.yona.nu/display/DEV/General+Requirements 
discussion ends saying to use Nickname. Not sure if it is applicable to notifications also.

Bert-R: Yes, it should be used everywhere. So then the nickname suffices (for buddies, for yourself it should be based on first and last name)

Sivateja-CMI: In notifications logged in user will never have to be shown.

Bert-R: They will: for goal conflicts of yourself.

Sivateja-CMI: Ok. In that case ```yona:user``` should be used to derive first name and last name.

Bert-R: :) Lucky you! You're right.

PR #523, opened 2018-10-10
APPDEV -1183 Wrong goal name in notification details

Sivateja-CMI:
Fixed notification details header with following fields in case of Nogo notifications.
1. Goal Name.
2. User Avatar name.
3. Yona header theme when clicked on notification and buddy profile inside it. 
    Blue for buddy and purple for self as per [wiki](https://wiki.yona.nu/display/DEV/Friend+-+Profile).


app/src/main/java/nu/yona/app/ui/message/NotificationFragment.java
```diff
@@ -75,137 +76,46 @@ public void onFriendsItemClick(View view)
...
+						handleGoalConflictMessageClick();
+						break;
+					case GOALCHANGEMESSAGE:
+						handleGoalChangeMessageClick();
+						break;
+					case BUDDYINFOCHANGEMESSAGE:
+						handleBuddyInfoChangeMessageClick();
+						break;
+					default:
+						messageIntentClicked = null;
```
Bert-R: Why would this need to be a member variable? It looks like another variable that should be passed over the stack instead, this time as return value of the various ``handleXxx`` methods, so you can pass it as input to ``replaceFragment`` below.

app/src/main/java/nu/yona/app/ui/message/NotificationFragment.java
```diff
@@ -75,137 +76,46 @@ public void onFriendsItemClick(View view)
 		{
 			if (view.getTag() instanceof YonaMessage)
 			{
-				YonaMessage yonaMessage = (YonaMessage) view.getTag();
-				Intent mMessageIntent = null;
-				if (yonaMessage.getNotificationMessageEnum() == NotificationMessageEnum.SYSTEM_MESSAGE)
+				yonaMessageClicked = (YonaMessage) view.getTag();
+				switch (yonaMessageClicked.getNotificationMessageEnum().getNotificationEnum())
```
Bert-R: Good to see this refactoring. It improves the readability a lot!

app/src/main/java/nu/yona/app/ui/message/NotificationFragment.java
```diff
@@ -244,12 +154,160 @@ public void onItemClick(View view)
...
+	{
+		try
+		{
+			SimpleDateFormat sdf = new SimpleDateFormat(AppConstant.YONA_LONG_DATE_FORMAT, Locale.getDefault());
+			Date date = sdf.parse(yonaMessageClicked.getActivityStartTime());
+			Calendar calendar = Calendar.getInstance();
+			calendar.setTime(date);
+			messageIntentClicked.putExtra(AppConstant.EVENT_TIME, calendar.get(Calendar.HOUR) + ":" + calendar.get(Calendar.MINUTE));
+		}
+		catch (Exception e)
```
Bert-R: Don't catch generic ``Exception`` but just catch what's needed (I know it was in the old code already, but please make this improvement along with the other improvements you're making).

app/src/main/java/nu/yona/app/ui/dashboard/SingleDayActivityDetailFragment.java
```diff
@@ -298,6 +301,17 @@ private void initilizeCommentControl(View view)
 	}
 
 
+	private void setUpDayAcvitivy()
+	{
+		if (getArguments() != null)
+		{
+			if (getArguments().get(AppConstant.OBJECT) instanceof DayActivity)
```
Bert-R: Don't nest if-statements unnecessarily. Nesting adds complexity.
As indicated in my earlier comment, this is sufficient:
```
if (getArguments() != null && getArguments().get(AppConstant.OBJECT) instanceof DayActivity)
```

PR #526, opened 2018-10-30
APPDEV-1181 Wrong buddy timeline goal time No-go.

Sivateja-CMI:
-- For No-go goal, time beyond goal is either zero or none. This is fixed. Application now shows the minutes beyond goal time.

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
+++ b/app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
@@ -80,42 +80,26 @@ public class TimelineStickyAdapter extends RecyclerView.Adapter<TimelineHolder> 
 	public void onBindViewHolder(final TimelineHolder holder, int position)
 	{
 		DayActivity dayActivity = (DayActivity) getItem(position);
+		holder.getView().setTag(dayActivity);
+		setUpTimeLineViewHolderWithDayActivity(holder, dayActivity);
+	}
 
+	private void setUpTimeLineViewHolderWithDayActivity(TimelineHolder holder, DayActivity dayActivity)
+	{
 		if (dayActivity != null)
 		{
 			switch (dayActivity.getChartTypeEnum())
 			{
 				case TIME_FRAME_CONTROL:
-					holder.getView().setTag(dayActivity);
-					if (dayActivity.getTimeZoneSpread() != null)
-					{
-						holder.getmTimeFrameGraph().chartValuePre(dayActivity.getTimeZoneSpread());
-					}
-					updateProfileImage(holder, dayActivity);
+					setUpTimeFrameControlHolderView(dayActivity, holder);
 					break;
 				case TIME_BUCKET_CONTROL:
-					holder.getView().setTag(dayActivity);
-					int maxDurationAllow = (int) dayActivity.getYonaGoal().getMaxDurationMinutes();
-					if (maxDurationAllow > 0)
-					{
-						holder.getmTimebucketGraph().graphArguments(dayActivity.getTotalMinutesBeyondGoal(), (int) dayActivity.getYonaGoal().getMaxDurationMinutes(), dayActivity.getTotalActivityDurationMinutes());
-					}
-					updateProfileImage(holder, dayActivity);
+					setUpTimeBucketControlHolderView(dayActivity, holder);
 					break;
 				case SPREAD_CONTROL:
 					break;
 				case NOGO_CONTROL:
-					holder.getView().setTag(dayActivity);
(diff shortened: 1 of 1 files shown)


app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
```diff
@@ -125,6 +109,40 @@ public void onBindViewHolder(final TimelineHolder holder, int position)
 		}
 	}
 
+
+	private void setUpTimeFrameControlHolderView(DayActivity dayActivity, TimelineHolder holder)
+	{
+		if (dayActivity.getTimeZoneSpread() != null)
```
Bert-R: Can this realistically be null?

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
```diff
@@ -125,6 +109,40 @@ public void onBindViewHolder(final TimelineHolder holder, int position)
...
+		{
+			holder.getmTimeFrameGraph().chartValuePre(dayActivity.getTimeZoneSpread());
+		}
+		updateProfileImage(holder, dayActivity);
+	}
+
+	private void setUpTimeBucketControlHolderView(DayActivity dayActivity, TimelineHolder holder)
+	{
+		int maxDurationAllow = (int) dayActivity.getYonaGoal().getMaxDurationMinutes();
+		if (maxDurationAllow > 0)
```
Bert-R: Any idea what this check is for? The max duration can never be negative. In case of a no-go goal, it is zero, but that you already classify as a no-go goal, so that won't come here.

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
```diff
@@ -125,6 +109,40 @@ public void onBindViewHolder(final TimelineHolder holder, int position)
...
+			holder.getmTimebucketGraph().graphArguments(dayActivity.getTotalMinutesBeyondGoal(), (int) dayActivity.getYonaGoal().getMaxDurationMinutes(), dayActivity.getTotalActivityDurationMinutes());
+		}
+		updateProfileImage(holder, dayActivity);
+	}
+
+	private void setUpNoGoHolderView(DayActivity dayActivity, TimelineHolder holder)
+	{
+		if (dayActivity.getGoalAccomplished())
+		{
+			holder.getmNogoImage().setImageResource(R.drawable.adult_happy);
```
Bert-R: Some day, you should rename these. It should be ``no_go_happy`` and ``no_go_sad``.

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
```diff
@@ -80,42 +80,26 @@ public TimelineHolder onCreateViewHolder(ViewGroup parent, int viewType)
 	public void onBindViewHolder(final TimelineHolder holder, int position)
 	{
 		DayActivity dayActivity = (DayActivity) getItem(position);
+		holder.getView().setTag(dayActivity);
+		setUpTimeLineViewHolderWithDayActivity(holder, dayActivity);
+	}
 
+	private void setUpTimeLineViewHolderWithDayActivity(TimelineHolder holder, DayActivity dayActivity)
+	{
 		if (dayActivity != null)
```
Bert-R: Can this realistically be null? The caller shouldn't call this method when it doesn't have a day activity.

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
Sivateja-CMI: I will note this in the 1.2 refactoring ticket

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
Sivateja-CMI: Done. Removed the null check.

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
Sivateja-CMI: Removed the check.

app/src/main/java/nu/yona/app/ui/frinends/TimelineStickyAdapter.java
Sivateja-CMI: No. Updated the code.

PR #528, opened 2018-10-30
APPDEV-1193 Profile Picture update triggers two notifications to buddies.

Sivateja-CMI:
-- Fixed by avoiding unnecessary Profile details update call when only profile pic is updated.

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/profile/EditDetailsProfileFragment.java
+++ b/app/src/main/java/nu/yona/app/ui/profile/EditDetailsProfileFragment.java
@@ -42,6 +42,7 @@ import nu.yona.app.api.manager.APIManager;
 import nu.yona.app.api.model.ErrorMessage;
 import nu.yona.app.api.model.Href;
 import nu.yona.app.api.model.RegisterUser;
+import nu.yona.app.api.model.User;
 import nu.yona.app.api.utils.ServerErrorCode;
 import nu.yona.app.customview.CustomAlertDialog;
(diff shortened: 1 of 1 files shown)


app/src/main/java/nu/yona/app/ui/profile/EditDetailsProfileFragment.java
```diff
@@ -254,13 +255,40 @@ public void onResume()
...
+		else
+		{
+			YonaActivity.getActivity().onBackPressed();
+		}
+	}
+
+	private boolean userDetailsChanged()
+	{
+		User loggedInUser = YonaApplication.getEventChangeManager().getDataState().getUser();
+		if (!firstName.getText().toString().equals(loggedInUser.getFirstName()))
```
Bert-R: Create a small helper method as follows:
```
private static boolean editTextEquals(EditText editText, String value)
{
	return editText.getText().toString().equals(value);
}
```

Now you can replace all of these if-statements with this:
```
return editTextEquals(firstName, loggedInUser.getFirstName()) ||
	editTextEquals(lastName, loggedInUser.getLastName()) ||
	editTextEquals(nickName, loggedInUser.getNickname()) ||
	editTextEquals(mobileNumber, loggedInUser.getMobileNumber());
```

Sivateja-CMI: Updated the code.

PR #531, opened 2018-11-27
APPDEV-1204 App activity monitoring fix drains battery

Sivateja-CMI:
-- App now has the Foreground service up and running forever, unless stopped by device in Doze mode.
-- After detecting doze mode broadcast, app schedules alarms for every 10 secs interval. Once alarm goes off and if device is interactive, app restarts the service else schedules next alarm.
-- PowerManager.ACTION_DEVICE_IDLE_MODE_CHANGED broadcast is received only when app goes to doze mode, but not when it comes out of it because of lack of active context.

### DIFF
--- a/app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
+++ b/app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
@@ -76,7 +76,7 @@ public class YonaReceiver extends BroadcastReceiver
 		PowerManager powerManager = (PowerManager) context.getSystemService(POWER_SERVICE);
 		if (powerManager.isDeviceIdleMode())
 		{
-			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
+			if (powerManager.isDeviceIdleMode() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
 			{
 				scheduleNextAlarmToCheckIfDeviceIsInteractive(context, INTERACTIVE_CHECK_INTERVAL);
 			}
--- a/app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
+++ b/app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
@@ -72,7 +72,7 @@ public class YonaReceiver extends BroadcastReceiver
 	@TargetApi(Build.VERSION_CODES.O)
 	private void handleDeviceDozeMode(Context context)
 	{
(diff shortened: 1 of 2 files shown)


app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
```diff
@@ -61,45 +61,59 @@ public void onReceive(Context context, Intent intent)
...
+				break;
 			default:
 				break;
 		}
 	}
 
+	@TargetApi(Build.VERSION_CODES.O)
+	private void handleDeviceDozeMode(Context context)
+	{
+		Logger.loge("BroadCast", "ACTION_DEVICE_IDLE_MODE_CHANGED");
```
Bert-R: Broadcast is a single word, so please spell with lowercase c

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
```diff
@@ -61,45 +61,59 @@ public void onReceive(Context context, Intent intent)
...
+		PowerManager powerManager = (PowerManager) context.getSystemService(POWER_SERVICE);
+		if (powerManager.isDeviceIdleMode())
+		{
+			if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
```
Bert-R: Combine with the above condition, like this:
```suggestion
			if (powerManager.isDeviceIdleMode() && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
```

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
```diff
@@ -76,7 +76,7 @@ private void handleDeviceDozeMode(Context context)
 		PowerManager powerManager = (PowerManager) context.getSystemService(POWER_SERVICE);
 		if (powerManager.isDeviceIdleMode())
```
Bert-R: The proposed changes feature is rather incomplete at the moment, as it doesn't allow doing a multiline change. You still need to remove this line and the related curly brace, and reduce the indentation.

Sivateja-CMI: yes. noticed and corrected in the update below.

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
Sivateja-CMI: done. update pushed.

PR #536, opened 2018-12-05
APPDEV-1197 vpn does not connect

Sivateja-CMI:
Fixed crash in log statement.


app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -598,19 +597,12 @@ public static boolean isVPNConnected(Context context)
 
 		String profileUUID = YonaApplication.getEventChangeManager().getSharedPreference().getUserPreferences().getString(PreferenceConstant.PROFILE_UUID, "");
 		VpnProfile profile = ProfileManager.get(context, profileUUID);
-		if (ProfileManager.getLastConnectedVpn() != null && profile != null)
-		{
-			Logger.loge("VPNCheck", "last connected vpn profile UUID: " + ProfileManager.getLastConnectedVpn().getUUIDString());
-			Logger.loge("VPNCheck", "current profile id: " + profile.getUUIDString());
-		}
 		if (VpnStatus.isVPNActive() && ProfileManager.getLastConnectedVpn() == profile)
```
Bert-R: Change the if/else into just this:
```
return (VpnStatus.isVPNActive() && ProfileManager.getLastConnectedVpn() == profile);
```

app/src/main/java/nu/yona/app/ui/YonaActivity.java
```diff
@@ -1575,7 +1575,6 @@ public void onClick(DialogInterface dialog, int which)
 
 	private void startVPN()
 	{
-		Logger.loge("VPNCheck", "VPN Initiated");
```
Bert-R: Good to see you reverted the logging.

app/src/main/java/nu/yona/app/utils/AppConstant.java
```diff
@@ -292,4 +292,5 @@
 	String RESTART_VPN = "com.yona.app.RESTART_VPN";
 	String RESTART_DEVICE = "com.yona.app.RESTART_DEVICE";
 	String WAKE_UP = "com.yona.app.WAKE_UP_ALARM";
+	String SHOW_VPN_WINDOW = "showVpnWindow";
```
Bert-R: Let's stick with the naming convention:
```
String SHOW_VPN_WINDOW = "com.yona.app.SHOW_VPN_WINDOW";
```

The naming convention might be unnecessary, but it's a pattern we have now and it doesn't cause any pain.

Sivateja-CMI: the pattern is only for broad cast actions, if you see the constants above the visible lines here, there is no pattern. I have placed in the wrong location in the file.

Bert-R: OK, fine. Can you move it to a more appropriate place, so that question won't be asked by the next reviewer/editor?

app/src/main/java/nu/yona/app/utils/AppUtils.java
Sivateja-CMI: I forgot to revert the changes I made earlier. Updated the code. Thanks.

PR #550, opened 2018-12-26
APPDEV-1222 global variables for preferences construct

Sivateja-CMI:
-- Added Global variables for App/user preferences and data state, which holds useful data across Application lifetime.
-- This is will remove lengthy constructs to access data from YonaApplication Class.

### DIFF
--- a/app/src/main/java/nu/yona/app/YonaApplication.java
+++ b/app/src/main/java/nu/yona/app/YonaApplication.java
@@ -9,6 +9,7 @@
 package nu.yona.app;
 
 import android.app.Application;
+import android.content.SharedPreferences;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.os.StrictMode;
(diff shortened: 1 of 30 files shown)


app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -36,6 +38,12 @@
 	private static EventChangeManager eventChangeManager;
 	private Tracker tracker;
 
+
+	public static SharedPreferences sharedAppPreferences;
```
Bert-R: Please wrap these three in accessor methods (e.g. ``getSharedAppPreferences()``) and make the variables private.  Modifiable public variables is considered a bad practice.

app/src/main/java/nu/yona/app/ui/message/NotificationFragment.java
```diff
@@ -399,7 +400,7 @@ private void refreshAdapter()
 		mMessageStickyRecyclerAdapter.clear();
 		currentPage = 0;
 		getUserMessages(false);
-		//YonaApplication.getEventChangeManager().getDataState().setEmbeddedWithBuddyActivity(null);
+		//sharedAppDataState.setEmbeddedWithBuddyActivity(null);
```
Bert-R: Remove this line of dead code

Sivateja-CMI: Done. Update pushed.

app/src/main/java/nu/yona/app/YonaApplication.java
Sivateja-CMI: Done. Update pushed.

PR #551, opened 2018-12-26
APPDEV-1209 comments are displayed twice in single day activity view

madhuvardan-CMI:
-Fixed issue
-Refactoring of DayActivityDetailFragment.java

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
+++ b/app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
@@ -11,6 +11,7 @@ package nu.yona.app.ui.dashboard;
 import android.content.Intent;
 import android.os.Bundle;
 import android.os.Handler;
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.content.ContextCompat;
 import android.support.v4.view.ViewPager;
@@ -42,7 +43,7 @@ import nu.yona.app.customview.YonaFontButton;
 import nu.yona.app.customview.YonaFontEditTextViewGeneral;
 import nu.yona.app.customview.YonaFontTextView;
 import nu.yona.app.enums.IntentEnum;
-import nu.yona.app.listener.DataLoadListener;
+import nu.yona.app.listener.DataLoadListenerImpl;
 import nu.yona.app.state.EventChangeListener;
 import nu.yona.app.state.EventChangeManager;
 import nu.yona.app.ui.BaseFragment;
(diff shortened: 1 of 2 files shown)


app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -98,148 +99,131 @@ public void onClick(View v)
...
 	@Nullable
 	@Override
-	public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState)
+	public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState)
+	{
+		initializeViewAttributes(inflater);
+		initializeCommentControl(view);
+		if (getArguments() != null)
+		{
+			if (getArguments().get(AppConstant.OBJECT) instanceof DayActivity)
```
Bert-R: Combine above if-statement with this one:
```
 			if (getArguments() != null && getArguments().get(AppConstant.OBJECT) instanceof DayActivity)
```

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -98,148 +99,131 @@ public void onClick(View v)
...
+			{
+				activity = (DayActivity) getArguments().get(AppConstant.OBJECT);
```
Bert-R: This argument lookup construct pretty common. I suggest adding the below two methods to ``BaseFragment``:
```
	protected <T> T getArgument(String key, Class<T> type)
	{
		return getArgument(key, type, null);
	}

	protected <T> T getArgument(String key, Class<T> type, T defaultValue)
	{
		Bundle arguments = getArguments();
		if (arguments == null)
		{
			return defaultValue;
		}
		Object argument = arguments.get(key);
		if (type.isInstance(argument))
		{
			return type.cast(argument);
		}
		return defaultValue;
	}
```
With that, the entire if-statement to assign the activity if given as argument can be replace with this:
```
activity = getArgument(AppConstant.OBJECT, DayActivity.class, activity);
```

In other places, you just want to get the argument value and null if it's not there. That's what the first method does:
```
activity = getArgument(AppConstant.OBJECT, DayActivity.class);
```

Note that the above construct can be used to replace calls to ``Object get(String key)``, but also to ``Serializable getSerializable(String key)`` and ``String getString(String key)``.

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -256,19 +240,47 @@ public void onPageSelected(int position)
...
-
 				}
 			});
 		}
+	}
 
-		YonaApplication.getEventChangeManager().registerListener(this);
-		setHook(new YonaAnalytics.BackHook(AnalyticsConstant.BACK_FROM_DAY_ACTIVITY_DETAIL_SCREEN));
-		return view;
+	private void previousDayActivity()
```
Bert-R: It's unclear what this method does. It's name sounds like a statement instead of an action (or a condition, in case of a boolean-returning ``isXxx`` method). It should be something like ``enablePreviousDayActivity``, ``goToPreviousDayActivity``, ``showPreviousDayActivity``, etc.

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -256,19 +240,47 @@ public void onPageSelected(int position)
...
+	{
+		if (viewPager.getCurrentItem() != 0)
+		{
+			YonaAnalytics.createTapEventWithCategory(AnalyticsConstant.DAY_ACTIVITY_DETAIL_SCREEN, AnalyticsConstant.PREVIOUS);
+			viewPager.setCurrentItem(viewPager.getCurrentItem() - 1);
+		}
+	}
+
+	private void nextDayActivity()
```
Bert-R: Same here.

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -256,19 +240,47 @@ public void onPageSelected(int position)
...
+		if (viewPager.getCurrentItem() != dayActivityList.size() - 1)
+		{
+			YonaAnalytics.createTapEventWithCategory(AnalyticsConstant.DAY_ACTIVITY_DETAIL_SCREEN, AnalyticsConstant.NEXT);
+			viewPager.setCurrentItem(viewPager.getCurrentItem() + 1);
+		}
+	}
+
+	private void setOnClickListenerSendButton()
+	{
+		if (!TextUtils.isEmpty(messageTxt.getText()))
```
Bert-R: Reverse and return:
```
 		if (TextUtils.isEmpty(messageTxt.getText()))
 		{
 		    return;
 		}
```

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -296,28 +308,26 @@ public void getCurrentDayActivityDetails(DayActivity dayActivity)
...
 		}
+	}
 
+	private Object handleDetailOfEachSpreadWithDayActivityFetchSuccess(Object result)
+	{
+		activity = (DayActivity) result;
+		isDataLoading = false;
+		YonaActivity.getActivity().showLoadingView(false, null);
+		setDayActivityDetails();
+		return null; // Dummy return value, to allow use as data load handler
```
Bert-R: 👍

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -326,131 +336,153 @@ private void setDayActivityDetails()
...
+		else
+		{
+			goBack();
+		}
+		setDayDetailTitleAndIcon();
+	}
 
+	private void setUpDayActivityListFromEmbeddedYonaActivity(EmbeddedYonaActivity embeddedYonaActivity)
+	{
+		try
```
Bert-R: This try/catch can be omitted, as there are no checked exceptions thrown from within this block that need to be caught.

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -475,36 +507,22 @@ private void updateFlow(int position)
 		}
 	}
 
+	//TODO
```
Bert-R: What needs to be done?

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -98,148 +99,131 @@ public void onClick(View v)
...
+	private void initializeOnClickListeners()
+	{
+		previousItem.setOnClickListener(view -> previousDayActivity());
+		nextItem.setOnClickListener(view -> nextDayActivity());
+		sendButton.setOnClickListener(view -> setOnClickListenerSendButton());
+	}
 
+	private void initializeOnPageClickListener()
+	{
 		if (viewPager != null)
```
Bert-R: Reverse and return:
```suggestion
 		if (viewPager == null)
 		{
 		    return;
 		}
```

app/src/main/java/nu/yona/app/ui/dashboard/DayActivityDetailFragment.java
```diff
@@ -296,28 +308,26 @@ public void getCurrentDayActivityDetails(DayActivity dayActivity)
 		{
```
Bert-R: Reverse and return:
```
if (isDataLoading)
{
    return;
}
```

madhuvardan-CMI:
I have made the changes and committed.

madhuvardan-CMI:
Applied the changes, will use this approach in future commits (refactoring).

PR #552, opened 2018-12-28
APPDEV-1210 comments are displayed twice in single week activity overview

madhuvardan-CMI:
-Fixed the issue
-Refactoring of WeekActivityDetailFragment.java

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
+++ b/app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
@@ -157,8 +157,8 @@ public class WeekActivityDetailFragment extends BaseFragment implements EventCha
 	public void onCreate(@Nullable Bundle savedInstanceState)
 	{
 		super.onCreate(savedInstanceState);
-		yonaBuddy = getArgument(AppConstant.YONA_BUDDY_OBJ, YonaBuddy.class, yonaBuddy);
-		if (yonaBuddy == null)
+		yonaBuddy = getArgument(AppConstant.YONA_BUDDY_OBJ, YonaBuddy.class);
+		if (yonaBuddy == null && getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class) != null)
 		{
 			yonaBuddy = APIManager.getInstance().getActivityManager().findYonaBuddy(getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class));
 		}
--- a/app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
+++ b/app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
@@ -348,10 +348,13 @@ public class WeekActivityDetailFragment extends BaseFragment implements EventCha
 				if (itemIndex >= 0)
 				{
 					customPageAdapter.notifyDataSetChanged(weekActivityList);
-					fetchComments(itemIndex);
 					if (itemIndex != viewPager.getCurrentItem())
 					{
 						viewPager.setCurrentItem(itemIndex);
+					}
+					else
+					{
+						fetchComments(itemIndex);
 					}
 					updateFlow(itemIndex);
 				}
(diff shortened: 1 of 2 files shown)


madhuvardan-CMI:
I have updated the code based on APPDEV-1209 PR.

app/src/main/java/nu/yona/app/ui/dashboard/WeekActivityDetailFragment.java
```diff
@@ -165,134 +157,133 @@ private void openDetailPage(Day activity)
 	public void onCreate(@Nullable Bundle savedInstanceState)
 	{
 		super.onCreate(savedInstanceState);
-		if (getArguments().get(AppConstant.YONA_BUDDY_OBJ) != null)
+		yonaBuddy = getArgument(AppConstant.YONA_BUDDY_OBJ, YonaBuddy.class, yonaBuddy);
```
Bert-R: The new behavior differs from what you had before. Is that desirable?

Earlier, it worked like this:
* If the argument ``YONA_BUDDY_OBJ`` is present:
  * If the argument is an instance of ``YonaBuddy``, use that
  * Else interpret it as an ``HRef`` and do a lookup

Now you have this:
* If the argument ``YONA_BUDDY_OBJ`` is present and it is an instance of ``YonaBuddy``, use that
* Else, use the current value of ``yonaBuddy``
* If that value is null, then interpret the argument as an ``HRef`` and do a lookup

Remarkable differences:
* You now can end up with the old value of ``yonaBuddy``, even though an ``HRef`` argument was provided
* You might be performing the lookup (with null) even though no argument was provided

If the argument can be missing, you'll have to add an enclosing if-statement and create something like this:
```
if (getArguments().get(AppConstant.YONA_BUDDY_OBJ) != null)
{
    yonaBuddy = getArgument(AppConstant.YONA_BUDDY_OBJ, YonaBuddy.class); // Note there is no default value
    if (yonaBuddy == null)
    {
        yonaBuddy = APIManager.getInstance().getActivityManager().findYonaBuddy(getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class));
    }
}
```

madhuvardan-CMI: In remarkable differences you have mentioned:

- `yonaBuddy` will be null because the mentioned code is in `onCreate`, when this method is called there will be no instance of `yonaBuddy` (i.e it will be null).

- Regarding the null lookup, Even if `getArguments()` is null `getArgument()` will return a null which leads to null lookup in `findYonaBuddy`. 

What I feel is either null check needs to be done for `getArguments()` too or else null check should be done for `HREF` in `findYonaBuddy`.

Bert-R: If indeed ``yonaBuddy`` is initially null in this context, then you can do it this way:
```
    yonaBuddy = getArgument(AppConstant.YONA_BUDDY_OBJ, YonaBuddy.class); // Note there is no default value
    if (yonaBuddy == null && getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class) != null)
    {
        yonaBuddy = APIManager.getInstance().getActivityManager().findYonaBuddy(getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class));
    }
```

Bert-R: > or else null check should be done for ``HREF`` in ``findYonaBuddy``.
This would not be a good idea. ``findYonaBuddy`` cannot do anything with null, so it shouldn't be called with it.

madhuvardan-CMI: I agree your point on `findYonaBuddy`. Final code should be like this
``` 
	if (getArguments() != null && getArguments().get(AppConstant.YONA_BUDDY_OBJ) != null)
		{
			yonaBuddy = getArgument(AppConstant.YONA_BUDDY_OBJ, YonaBuddy.class); // Note there is no default value
			if (yonaBuddy == null && getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class) != null)
			{
				yonaBuddy = APIManager.getInstance().getActivityManager().findYonaBuddy(getArgument(AppConstant.YONA_BUDDY_OBJ, Href.class));
			}
		}
```

Bert-R: What does the outer if-statement add? I believe we can do without it.

madhuvardan-CMI: Yeah agreed, no need of it. Will push the commit now

PR #553, opened 2019-01-07
APPDEV-1230 Fabric Fatal Exceptions on Build-13(175)

madhuvardan-CMI:
- Fixed the issue
- Now the alert for enabling the notification permission will shown on app launch

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/YonaActivity.java
+++ b/app/src/main/java/nu/yona/app/ui/YonaActivity.java
@@ -352,22 +352,10 @@ public class YonaActivity extends BaseActivity implements FragmentManager.OnBack
 		if (!NotificationManagerCompat.from(this).areNotificationsEnabled())
 		{
 			Logger.loge("Notifications Disabled", this.getString(R.string.notification_disabled_message));
-			displayErrorAlert(new ErrorMessage(this.getString(R.string.notification_disabled_message)));
+			AppUtils.displayErrorAlert(new ErrorMessage(this.getString(R.string.notification_disabled_message)));
 		}
 	}
 
-	public static void displayErrorAlert(ErrorMessage errorMessage)
-	{
-		AppUtils.runOnUiThread(() -> getGenericAlertDialogWithErrorMessage(errorMessage).show());
-	}
-
-	public static AlertDialog getGenericAlertDialogWithErrorMessage(ErrorMessage errorMessage)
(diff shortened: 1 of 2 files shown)


app/src/main/java/nu/yona/app/ui/YonaActivity.java
```diff
@@ -342,9 +343,31 @@ public void onResume()
...
+	private void checkForNotificationPermission()
+	{
+		if (!NotificationManagerCompat.from(this).areNotificationsEnabled())
+		{
+			Logger.loge("Notifications Disabled", this.getString(R.string.notification_disabled_message));
+			displayErrorAlert(new ErrorMessage(this.getString(R.string.notification_disabled_message)));
+		}
+	}
+
+	public static void displayErrorAlert(ErrorMessage errorMessage)
```
Bert-R: Why are you pulling this one and the next one out of ``AppUtils.java``? These methods are generic utility methods, so I favor leaving them there.

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -202,9 +201,7 @@ public static void startForegroundService(Context context, Intent activityMonito
 	{
 		if (!NotificationManagerCompat.from(context).areNotificationsEnabled())
 		{
-			Logger.loge("Notifications Disabled", context.getString(R.string.notification_disabled_message));
-			displayErrorAlert(context, new ErrorMessage(context.getString(R.string.notification_disabled_message)));
-			return;
+			return; // Notification is required for starting a ForegroundService
```
Bert-R: Please change into "Notification *permission* is..."

madhuvardan-CMI:
- It occurs when user disables the notifications from Yona. If they disable notifications, app monitoring notification will not be there hence application will not monitor the usage. Earlier we used to display an alert while starting the service, where there wont be an activity to display the alert (if user removes the application from recents menu). Now we will be displaying the alert upon app launch.
- This crash is subjective to android 8.0 and above, only if user disables the notification for app.

Bert-R:
Thanks, clear!
Should we bring this to the 1.1 version too?

madhuvardan-CMI:
Even though it is a crash because we are handling it, fabric is reporting them as Non Fatal Exceptions. This fix can be included in 1.2 version.

PR #554, opened 2019-01-09
APPDEV-1114 Take device specific links from device instead of user

madhuvardan-CMI:
Update for application to take device specific links from embedded object of user. For following links
1. postOpenAppEvent
2. appActivity
3. sslRootCert
4. sslRootCertCN


app/src/main/java/nu/yona/app/api/model/EmbeddedYonaDevices.java
```diff
@@ -0,0 +1,51 @@
...
+	 * @return The yonaDevices
+	 */
+	public List<YonaDevice> getYonaDevices()
+	{
+		return yonaDevices;
+	}
+
+	public YonaDevice getCurrentDevice()
+	{
+		if (this.yonaDevices.size() == 1)
```
Bert-R: Please simplify this to just:
```
		for (YonaDevice device : yonaDevices)
		{
			if (device.isRequestingDevice())
			{
				return device;
			}
		}
		throw new IllegalStateException("Current device not found. Number of devices: " + yonaDevices.size());
```

There is no point in trying to move in with invalid data. The current code could make you use an invalid device.

app/src/main/java/nu/yona/app/api/model/User.java
```diff
@@ -257,12 +247,7 @@ public ContentValues getDbContentValues()
 
 	public String getSslRootCertCN()
 	{
-		return this.sslRootCertCN;
-	}
-
-	public void setSslRootCertCN(String sslRootCertCN)
-	{
-		this.sslRootCertCN = sslRootCertCN;
+		return this.getEmbedded().getYonaDevices().getEmbedded().getCurrentDevice().getSslRootCertCN();
```
Bert-R: The part ``this.getEmbedded().getYonaDevices().getEmbedded().getCurrentDevice()`` occurs four times. Just put it in a private method.

app/src/main/java/nu/yona/app/ui/pincode/BasePasscodeActivity.java
```diff
@@ -568,16 +567,14 @@ private void displayData()
 	protected void postOpenAppEvent()
 	{
 		User loggedInUser = getSharedAppDataState().getUser();
-		if (loggedInUser.getLinks() == null || loggedInUser.getLinks().getYonaPostOpenAppEvent() == null)
+		if (loggedInUser.getEmbedded() == null || loggedInUser.getEmbedded().getYonaDevices() == null)
```
Bert-R: Please put this behavior inside the user entity, exposed through a method like ``isDeleted()``.

Next, I don't understand this check. When there is no ``embedded`` or no devices inside it, it could very well indicate that the mobile number isn't confirmed yet. In that case, the account still exists.

What are you doing when you get a 404 from the server? Can't you put an explicit marker ``isDeleted`` in the entity?

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -626,37 +627,35 @@ private static Intent getVPNIntent(VpnProfile profile, Context context)
 	public static void downloadCertificates()
 	{
 		User user = YonaApplication.getEventChangeManager().getDataState().getUser();
-		if (user != null && user.getLinks() != null)
+		if (user == null || user.getEmbedded() == null || user.getEmbedded().getYonaDevices() == null
```
Bert-R: What does this do? IMHO, the check should be: ``if (!user.isMobileNumberConfirmed())``.
The other checks hide inconsistencies.

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -626,37 +627,35 @@ private static Intent getVPNIntent(VpnProfile profile, Context context)
...
-						if (trialCertificateCount < 3)
-						{
-							downloadCertificates();
-						}
-					}
-				});
-			}
+	private static Object handleDownloadFileFromUrlSuccess(Object result)
+	{
+		if (result != null && !TextUtils.isEmpty(result.toString()))
```
Bert-R: See [the documentation](https://developer.android.com/reference/android/text/TextUtils.html#isEmpty(java.lang.CharSequence)). This suffices:

```
		if (!TextUtils.isEmpty(result.toString()))
```

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -626,37 +627,35 @@ private static Intent getVPNIntent(VpnProfile profile, Context context)
...
+			YonaApplication.getEventChangeManager().notifyChange(EventChangeManager.EVENT_ROOT_CERTIFICATE_DOWNLOADED, null);
+		}
+		logi(TAG, "Download successful: " + result.toString());
+		return null; // Dummy return value, to allow use as data load handler
+	}
+
+	private static Object handleDownloadFileFromUrlFailure()
+	{
+		loge(TAG, "Download fail");
+		trialCertificateCount++;
```
Bert-R: Can you also rename this? We don't have trial certificates :)

I suggest naming it ``certificateDownloadAttempts``.

app/src/main/java/nu/yona/app/ui/pincode/BasePasscodeActivity.java
Sivateja-CMI: We have added this null check to avoid a crash in a rare scenario. I don't remember that now.
Getting a 404 from server and this is different right?

Sivateja-CMI: I can add `isDeleted` to the user entity with same check happening there.

Bert-R: The server will not have ``_embedded`` when the mobile number is not confirmed.

I propose to add a status value to the ``user`` entity, with three values:
* ``ACTIVE`` In this case, the user entity was successfully retrieved from the server and the ``yona:messages`` link is available.
* ``NUMBER_NOT_CONFIRMED`` In this case, the user entity was successfully retrieved from the server but the ``yona:messages`` link is NOT available.
* ``NOT_REGISTERED`` In this case, there is either no user entity at all (initial situation) or the server returned a 404 when trying to fetch it.

If that is available, you can rewrite statements like this as follows:
```
		if (loggedInUser.getStatus() != UserStatus.ACTIVE)
```

Bert-R: See also [APPDEV-1239](https://jira.yona.nu/browse/APPDEV-1239).

Sivateja-CMI: `NOT_REGISTERED` when does this state occur? I am little confused. 
If server returns 404 it can be that either URL is wrong or either API is wrong. So in this case user entity itself will not be there.
Once user response is returned from server, app will instantiate the entity in application, then the `self` url returned from the server will guide to fetch the user details further. 
For some reason if api returns 404 what is the purpose of maintaining the state, as we cannot recover the data anyway.

Bert-R: It's unclear to me how the app registers the state that there is no user account (you can get into that state through a fresh app installation and by removing your account). If that is tracked by not having the user entity, that's fine with me. Then you still have three states, but marked differently:

Two are reflected through the enum status value:
* ``ACTIVE`` In this case, the user entity was successfully retrieved from the server and the ``yona:messages`` link is available.
* ``NUMBER_NOT_CONFIRMED`` In this case, the user entity was successfully retrieved from the server but the ``yona:messages`` link is NOT available.

The third state ("not registered") is indicated by the fact that the logged in user is null.

app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -87,7 +87,8 @@
 	private static Intent activityMonitorIntent;
 	private static ScheduledExecutorService scheduler;
 	private static final YonaReceiver receiver = new YonaReceiver();
-	private static int trialCertificateCount = 0, trialVPNCount = 0;
+	private static int certificateDownloadAttempts = 0;
+	private static int vpnConnectionAttempts = 0;
```
Bert-R: Thanks for taking this one along

PR #556, opened 2019-01-11
APPDEV-1230 fabric fatal exceptions on build 13(175)

madhuvardan-CMI:
-Application now shows an alert for turning off persistent notifications.
-Update the alert message.


app/src/main/java/nu/yona/app/utils/AppUtils.java
```diff
@@ -398,6 +399,14 @@ private static AlertDialog getGenericAlertDialogWithErrorMessage(ErrorMessage er
 		return alertDialogBuilder.create();
 	}
 
+	@TargetApi(Build.VERSION_CODES.O)
+	public static boolean arePersistentNotificationsEnabled(Context context, String channelId)
+	{
+		android.app.NotificationManager notificationManager = context.getSystemService(android.app.NotificationManager.class);
+		NotificationChannel notificationChannel = notificationManager.getNotificationChannel(channelId);
+		return notificationChannel != null && notificationChannel.getImportance() == android.app.NotificationManager.IMPORTANCE_NONE;
```
Bert-R: ``notificationChannel`` can only be null when an invalid ``channelId`` is given, right? If that's true, then this condition should be changed into:
```
		return notificationChannel.getImportance() == android.app.NotificationManager.IMPORTANCE_NONE;
```
It will cause an NPE when the ID is wrong, thus exposing that issue.

app/src/main/java/nu/yona/app/ui/YonaActivity.java
```diff
@@ -349,11 +349,12 @@ public void onResume()
 
 	private void checkForNotificationPermission()
 	{
-		if (!NotificationManagerCompat.from(this).areNotificationsEnabled())
+		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O || (NotificationManagerCompat.from(this).areNotificationsEnabled() && !AppUtils.arePersistentNotificationsEnabled(getActivity(), AppConstant.YONA_SERVICE_CHANNEL_ID)))
```
Bert-R: Is this right?
It now returns when:
* Android is older than O
* The notifications are enabled and the persistent notifications are NOT enabled

That sounds wrong to me. I would expect this:
It now returns when:
* Android is older than O
* The notifications, including our persistent one, are enabled

madhuvardan-CMI: Functionality wise it was working fine, problem was with the way I'm using those conditions.
I have rectified them now

Bert-R: Yes, I see what you mean now.

app/src/main/java/nu/yona/app/ui/YonaActivity.java
```diff
@@ -349,11 +349,12 @@ public void onResume()
 
 	private void checkForNotificationPermission()
 	{
-		if (!NotificationManagerCompat.from(this).areNotificationsEnabled())
+		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O || (NotificationManagerCompat.from(this).areNotificationsEnabled() && AppUtils.arePersistentNotificationsEnabled(getActivity(), AppConstant.YONA_SERVICE_CHANNEL_ID)))
```
Bert-R: I should have asked in the first review round: are you sure the persistent notification needs to be enabled for Yona to work properly?

madhuvardan-CMI: Yes, from oreo and above without a persistent notification activity monitoring service doesn't work.

Bert-R: In that case, please make a PR for ``master`` too with the same change set. For the error message, I felt it wasn't worth it, but this part makes it worthwhile.

app/src/main/java/nu/yona/app/utils/AppUtils.java
Sivateja-CMI: At the very beginning of the application, the notificationChannel can be null on the first install of the application, which will cause a NPE here. So I believe null check is needed. I will test the hotfix branch and confirm it.

Bert-R: 👍

Sivateja-CMI: It causes a crash on first install on my device. As the channel with said ID is not yet created by this time. I will add the check back.

Bert-R: 👍 
Please add a comment explaining the reason, so we don't remove it again as unnecessary.

Sivateja-CMI: Adding this null check creates another issue. This will cause alert to popup even on first launch even before user is aware of what the channel is about. 
I believe silently creating the channel is better solution. And use this channel when needed with the ID.

Bert-R: That makes sense indeed.

PR #558, opened 2019-01-15
APPDEV-1236 Changes from appdev1230.

Sivateja-CMI:
-- Check for notifications access for persistent notifiations.
-- Minor code fixes for the update of new channel id's and other issues.


app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
```diff
@@ -152,34 +154,38 @@ private void startService(Context context)
...
-			NotificationManager notificationManager = (NotificationManager) context.getSystemService(context.NOTIFICATION_SERVICE);
-			notificationManager.notify(0, notification);
+			return;
 		}
-		catch (Exception e)
+		NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
 		{
-			AppUtils.reportException(YonaReceiver.class.getSimpleName(), e, Thread.currentThread());
+			NotificationChannel channel = new NotificationChannel(AppConstant.YONA_VPN_CHANNEL_ID,
```
Bert-R: Do we really need to take this in? In the 1.2 version of the app, it doesn't work very well:
* "Click to connect again" opens the application instead of just connecting the VPN again
* Upon opening the app, it connects the VPN again but doesn't remove the "VPN disconnected" notification

So if we don't really have to, I'd rather leave out the VPN notification from the 1.1 release.

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
```diff
@@ -152,34 +154,38 @@ private void startService(Context context)
 	}
 
 
-	private void showRestartVPN(final String message)
+	private void showRestartVPN(String message)
 	{
-		try
+		Intent intent = new Intent(context.getApplicationContext(), YonaActivity.class);
+		PendingIntent pendingIntent = PendingIntent.getActivity(context, 0, intent, 0);
+		if (pendingIntent == null)
```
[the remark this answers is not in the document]
Sivateja-CMI: Actually the pending intent here is created by the VPN framework. So I thought of not taking chance. And there was also a [exception](https://fabric.io/yonadev/android/apps/nu.yona.app/issues/5c0982d2f8b88c2963d15d81/sessions/latest) registered in the Fabric in the same place.

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
Sivateja-CMI: Agree, with the issue, that It will not remove the notification after VPN connection. 
We can ignore the broadcast for 1.1 and correct the behaviour in 1.2.

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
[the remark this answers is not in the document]
Bert-R: The documentation clearly states the return value cannot be null unless ``FLAG_NO_CREATE`` is given, so there is no need to test the return value for null.

The exception in Fabric is an NPE _inside_ ``PendingIntent.getActivity``.  The parameter ``intent``  was likely null. Check [the code for build 175](https://github.com/yonadev/yona-app-android/blob/aa2ff954f1be8f519104d71fcf5a31360c70d876/app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java):
```
			Intent intent = AppUtils.startVPN(context, true);
			PendingIntent pIntent = PendingIntent.getActivity(context, (int) System.currentTimeMillis() + 10000, intent, 0);
```

The implementation of ``AppUtils.startVPN`` has two code paths were it can return null. See the implementation [here](https://github.com/yonadev/yona-app-android/blob/aa2ff954f1be8f519104d71fcf5a31360c70d876/app/src/main/java/nu/yona/app/utils/AppUtils.java).

app/src/main/java/nu/yona/app/api/receiver/YonaReceiver.java
Bert-R: OK

app/src/main/java/nu/yona/app/utils/AppConstant.java
```diff
@@ -295,4 +295,9 @@
 	String RESTART_VPN = "com.yona.app.RESTART_VPN";
 	String RESTART_DEVICE = "com.yona.app.RESTART_DEVICE";
 	String WAKE_UP = "com.yona.app.WAKE_UP_ALARM";
+
+	// Notifications Channel Id's
+	String YONA_VPN_CHANNEL_ID = "yona-vpn-channel";
```
Bert-R: This one should be left out now.

app/src/main/res/values/strings.xml
```diff
@@ -226,7 +226,8 @@
     <string name="generic_exception_message">Unknown Exception. Please contact Yona support</string>
     <string name="generic_mobile_number_format_error">Invalid Mobile Number. Please check the format</string>
     <string name="yona_notification_content">Monitoring app usage</string>
-    <string name="notification_disabled_message">Notifications for the Yona application have been disabled by the user. Give permission in device settings.</string>
-    <string name="notification_channel_name">Persistent notifications</string>
+    <string name="notification_disabled_message">Yona cannot monitor app activity because notifications for the Yona application have been disabled. Enable notifications for Yona again in the device settings.</string>
+    <string name="yona_service_notification_channel_name">Persistent notifications</string>
+    <string name="yona_vpn_notification_channel_name">VPN notifications</string>
```
Bert-R: This one can be left out now.

Sivateja-CMI: Thanks for highlighting. Added back the missing translations.

app/src/main/java/nu/yona/app/utils/AppConstant.java
Sivateja-CMI: Done.

PR #559, opened 2019-01-16
APPDEV-1239 lost account. Issue since APPDEV-1222

Sivateja-CMI:
-- Corrected the Preferences Initialization which went wrong in previous commits. Due to which App was unable to recognize the already logged in user. 
This bug was introduced in APPDEV-1222.

### DIFF
--- a/app/src/main/java/nu/yona/app/YonaApplication.java
+++ b/app/src/main/java/nu/yona/app/YonaApplication.java
@@ -109,10 +109,10 @@ public class YonaApplication extends Application
 
 	private void validateAndInitializePreferences()
 	{
-		sharedAppPreferences = eventChangeManager.getSharedPreference().getUserPreferences();
-		sharedUserPreferences = eventChangeManager.getSharedPreference().getAppPreferences();
+		sharedAppPreferences = eventChangeManager.getSharedPreference().getAppPreferences();
+		sharedUserPreferences = eventChangeManager.getSharedPreference().getUserPreferences();
 		if (sharedUserPreferences.getString(AppConstant.SERVER_URL, null) != null)
-		{
+		{  // Need to exchange the preferences, as a couple of previous builds had wrong initializations.
 			SharedPreferences tempSharedUserPrefs = YonaApplication.getAppContext().getSharedPreferences("TEMP_USER_PREF", Context.MODE_PRIVATE);
 			AppUtils.copySharedPreferences(sharedAppPreferences, tempSharedUserPrefs, true);
 			SharedPreferences tempSharedAppPrefs = YonaApplication.getAppContext().getSharedPreferences("TEMP_APP_PREF", Context.MODE_PRIVATE);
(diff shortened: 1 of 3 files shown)


app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -100,12 +102,27 @@ public void onCreate()
...
 		Foreground.init(this);
 	}
 
+	private void validateAndInitializePreferences()
+	{
+		sharedAppPreferences = eventChangeManager.getSharedPreference().getAppPreferences();
+		sharedUserPreferences = eventChangeManager.getSharedPreference().getUserPreferences();
+		if (sharedUserPreferences.getString(AppConstant.SERVER_URL, null) != null)
+		{  // Need to exchange the preferences as corrective action, as a couple of previous builds had wrong(Exchanged Preferences) initializations.
+			nu.yona.app.utils.Logger.logi("Preferences", "Preferences Exchanged");
```
Bert-R: This log message says "exchanged" (i.e. it's done now) while it's yet to exchange the preferences.
I see two options:
* Move this line to just below 122
* Copy this line to just below 122 and change this one into "Exchanging preferences"

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -100,12 +102,27 @@ public void onCreate()
...
+			SharedPreferences tempSharedUserPrefs = YonaApplication.getAppContext().getSharedPreferences("TEMP_USER_PREF", Context.MODE_PRIVATE);
+			AppUtils.copySharedPreferences(sharedAppPreferences, tempSharedUserPrefs, true);
```
Bert-R: You're always calling it with ``true``, right? If so, please omit the parameter and call the method ``moveSharedPreferences``, as you're practically moving them from one place to another.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -100,12 +102,27 @@ public void onCreate()
...
 		sharedAppDataState = eventChangeManager.getDataState();
 		Foreground.init(this);
 	}
 
+	private void validateAndInitializePreferences()
+	{
+		sharedAppPreferences = eventChangeManager.getSharedPreference().getAppPreferences();
+		sharedUserPreferences = eventChangeManager.getSharedPreference().getUserPreferences();
+		if (sharedUserPreferences.getString(AppConstant.SERVER_URL, null) != null)
+		{  // Need to exchange the preferences as corrective action, as a couple of previous builds had wrong(Exchanged Preferences) initializations.
```
Bert-R: It's unclear to me what this code block does. The comment doesn't help me (don't understand what "Exchanged Preferences") and I don't understand why the code below does something useful. When I'm not mistaken, it moves the preferences to a temp location and back.

Please enlighten me

Sivateja-CMI: `sharedAppPreferences = eventChangeManager.getSharedPreference().getUserPreferences();
		sharedUserPreferences = eventChangeManager.getSharedPreference().getAppPreferences();`

There were the statements earlier. App Preferences were copied into userPreferences variable and viceversa. This was causing the app to get logged out as it does not find logged in user.
Current code will find if preferences are wrongly initialized or not. If yes, they will be exchanged otherwise will continue. 
Exchanging is needed if user comes from previous build versions where wrong code is used.

app/src/main/java/nu/yona/app/YonaApplication.java
Sivateja-CMI: Done.

app/src/main/java/nu/yona/app/YonaApplication.java
Sivateja-CMI: Done.

PR #569, opened 2019-01-29
APPDEV-1251 Close app if mandatory network connection is unavailable

madhuvardan-CMI:
Application will now close upon clicking "ok" button in alert dialog which is displayed when there is no network available at the Launch Activity

### DIFF
--- a/app/src/main/java/nu/yona/app/ui/LaunchActivity.java
+++ b/app/src/main/java/nu/yona/app/ui/LaunchActivity.java
@@ -141,10 +141,16 @@ public class LaunchActivity extends BaseActivity
 	{
 		if (!AppUtils.isNetworkAvailable(YonaApplication.getAppContext()))
 		{
-			AppUtils.displayErrorAlert(this, new ErrorMessage(this.getString(R.string.mandatory_user_fetch_failure)));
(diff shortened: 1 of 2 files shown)


app/src/main/java/nu/yona/app/ui/LaunchActivity.java
```diff
@@ -141,12 +141,18 @@ private void reloadUserFromServer(String url)
...
 		if (!AppUtils.isNetworkAvailable(YonaApplication.getAppContext()))
 		{
-			AppUtils.displayErrorAlert(this, new ErrorMessage(this.getString(R.string.mandatory_user_fetch_failure)));
+			AppUtils.displayErrorAlert(this, new ErrorMessage(this.getString(R.string.mandatory_user_fetch_failure)), this::closeApplicationIfNoNetwork);
 			return;
 		}
 		getUserFromServer(url);
 	}
 
+	public void closeApplicationIfNoNetwork(DialogInterface dialog, int which)
```
Bert-R: This name would be right if the method would close the application if there is no network, but it closes it always, irrespective of the network connectivity. So just call it ``closeApplication``.

Bert-R: This method can be private

PR #572, opened 2019-02-01
APPDEV-1256 Crash in Per day/week fragments due to http scheme issue.

Sivateja-CMI:
Cause : App user is logged into to https+domain and stored user object has http+domain. Which cause goals self links mismatch and runtime crash.
Fix:
-- App compares stored user self link with environment base url logged in. If url schemes are different then app defaults the user with environment scheme.
-- Reloads the user data with corrected url and stores locally.


app/src/main/java/nu/yona/app/ui/LaunchActivity.java
```diff
@@ -117,40 +118,86 @@ else if (!sharedUserPreferences.getBoolean(PreferenceConstant.STEP_PASSCODE, fal
...
+		if (!AppUtils.isNetworkAvailable(YonaApplication.getAppContext()))
 		{
-			@Override
-			public boolean onLongClick(View v)
-			{
-				switchEnvironment();
-				return true;
-			}
+			AppUtils.displayErrorAlert(this, new ErrorMessage(this.getString(R.string.mandatory_user_fetch_failure)));
+			return;
```
Bert-R: The app should exit here. Otherwise, the user can go through the sign-up sequence, thus loosing their account.

PR #573, opened 2019-02-04
APPDEV-1110 take the vpn profile from the device entity rather than the user

madhuvardan-CMI:
-Updated the code to take vpn profile from the device entity rather than the user
-Added a getter for user object from local db.
-Minor refactoring of AppUtils.java


app/src/main/java/nu/yona/app/api/model/YonaDevice.java
```diff
@@ -16,6 +16,10 @@
 	@Expose
 	private String sslRootCertCN;
 
+	@SerializedName("vpnProfile")
+	@Expose
+	private VpnProfile vpnProfile;
```
Bert-R: You need to increment the version number of the user data, as older versions don't have the VPN profile.

PR #577, opened 2019-02-08
APPDEV-1255 Fix for buddy timeline crash issue while getting the buddy goals from stored user object.

Sivateja-CMI:
-- Added get user call in timeline to get updated the user with new buddies and their goals.
-- Minor refactoring.


app/src/main/java/nu/yona/app/api/manager/impl/ActivityManagerImpl.java
```diff
@@ -1148,19 +1141,19 @@ public YonaBuddy findYonaBuddy(Href yonaBuddy)
...
 	{
 		User user = getAppUser();
-		if (!isUserWithBuddies(user))
-		{
-			return null;
-		}
 		List<YonaBuddy> yonaBuddies = user.getBuddies();
 		for (YonaBuddy buddy : yonaBuddies)
 		{
 			if (buddy.getYonaGoals() != null)
```
Bert-R: This never returns null, so the condition can be removed.

app/src/main/java/nu/yona/app/api/manager/impl/ActivityManagerImpl.java
```diff
@@ -1148,19 +1141,19 @@ public YonaBuddy findYonaBuddy(Href yonaBuddy)
...
-				return findMatchingGoalFromYonaBuddy(goalHref, buddy);
+				YonaGoal matchingGoal = findMatchingGoalFromYonaBuddy(goalHref, buddy);
+				if (matchingGoal != null)
+				{
+					return matchingGoal;
+				}
 			}
 		}
-		return null; // Dummy return value, to allow use as data load handler
+		return null;
```
Bert-R: This is invalid. If we can't find the goal, something is badly wrong and it needs to be flagged right here. Please throw an exception like this:
```
throw new IllegalStateException("Buddy goal not found: " + goalHref.getHref());
```

app/src/main/java/nu/yona/app/api/manager/impl/ActivityManagerImpl.java
```diff
@@ -1425,7 +1410,7 @@ private void filterAndUpdateWithBuddyData(EmbeddedYonaActivity embeddedYonaActiv
 				listener.onError(new ErrorMessage(mContext.getString(R.string.no_data_found)));
 			}
 		}
-		catch (NullPointerException e)
+		catch (Exception e)
```
Bert-R: As we're not expecting any exception here, don't catch.

app/src/main/java/nu/yona/app/ui/BaseActivity.java
```diff
@@ -34,35 +32,32 @@
...
 
 	/**
 	 * Show loading view.
 	 *
 	 * @param loading the loading
 	 * @param message the message
 	 */
-	public void showLoadingView(boolean loading, String message)
+
+	public void toggleLoadingView(boolean loading, String message)
```
Bert-R: The method name is better, but it's still wrong. Toggle suggests you're hiding if it it's visible and showing it when it's not visible.

Given that you've removed all logic (which was a good action!), just change the callers: make then call ``displayLoadingView`` or ``dismissLoadingView`` depending on their needs.

app/src/main/java/nu/yona/app/ui/LaunchActivity.java
```diff
@@ -131,7 +131,7 @@ private void ensureValidUserEntity()
 			User user = APIManager.getInstance().getAuthenticateManager().getUser();
 			URL environmentURL = new URL(YonaApplication.getEventChangeManager().getDataState().getServerUrl());
 			URL storedUserURL = new URL(user.getLinks().getSelf().getHref());
-			if (environmentURL.getProtocol() != storedUserURL.getProtocol())
+			if (!environmentURL.getProtocol().equals(storedUserURL.getProtocol()))
```
Bert-R: Oops! I didn't see this one. Is it wrong in the ``master`` too?

app/src/main/java/nu/yona/app/ui/BaseActivity.java
```diff
@@ -34,35 +32,32 @@
...
+				displayLoadingView();
 			}
 			else if (progressDialog != null && !loading)
 			{
-				progressDialog.dismiss();
-				progressDialog = null;
+				dismissLoadingView();
 			}
 		}
 		catch (Exception e)
```
Bert-R: As we're not expecting an exception here, we shouldn't be catching it. Higher layers will take care of it.

app/src/main/java/nu/yona/app/ui/BaseActivity.java
Bert-R: The parameter ``message`` isn't used, so remove it.

app/src/main/java/nu/yona/app/ui/LaunchActivity.java
Sivateja-CMI: Yes it is in master too.

Bert-R: In that case, can you make a PR on master for this?

Sivateja-CMI: But I doubt if there will be any harm with it. As both the references will never be equal and app will make an extra getUser call.

Bert-R: But that'll cause extra overhead to everyone. It's better to make that 8-character PR. Pushing that to production is a very small effort.

Sivateja-CMI: ok.

PR #579, opened 2019-02-11
APPDEV-1240 disable back action in pin reset screen

madhuvardan-CMI:
-disabled back action in pin reset
-updated the code to show correct timer at pin reset screen.
-refactoring of BasePasscodeActivity.java
-added an alert upon clicking pin reset button

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -73,7 +73,7 @@
     <string name="inloggin">LOGIN</string>
     <string name="enterpasscode">You need to enter a 6 digit alphanumeric passcode.</string>
     <string name="resetpinrequest">Your request for reset has been received. You will receive an text message with a PIN reset confirmation code within %1$s.</string>
-    <string name="resetpinalert">Do you want to reset your pin?</string>
+    <string name="resetpinalert">Do you really want to reset your PIN? For security reasons, that will take time and you cannot use Yona in the meanwhile.</string>
     <string name="useroverride">Mobile Number %1$s is already registered with Yona. Do you want to re-register same number? (Note: All data associated with this number will be removed, if you want to add another device, you can go back and click on Login.)</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -73,6 +73,7 @@
     <string name="inloggin">LOGIN</string>
     <string name="enterpasscode">You need to enter a 6 digit alphanumeric passcode.</string>
     <string name="resetpinrequest">Your request for reset has been received. You will receive an text message with a PIN reset confirmation code within %1$s.</string>
+    <string name="resetpinalert">Do you want to reset your pin?</string>
```
madhuvardan-CMI: can you please provide the translated string

Bert-R: Please change this text into:
> Do you really want to reset your PIN? For security reasons, that will take time and you cannot use Yona in the meanwhile.

For Dutch:
> Wil je echt je pincode resetten? Om veiligheidsredenen gaat daar tijd overheen en ondertussen kan je Yona een niet gebruiken.

madhuvardan-CMI:
If a user decides to reset his pin and clicks on it, then we will show a timer (which shows how much it will take to get the pin reset code. If the user force closes the app, once after restarting it he will end up at pin reset screen itself. Currently app is behaving like this, is this what it meant to be or any other behaviour?

Bert-R:
>If the user force closes the app, once after restarting it he will end up at pin reset screen itself.

Is that the screen to request a PIN reset? In that case, it's wrong, as the reset request is already sent and some of the time has passed by already.

This is not right, but let's not invest the time in fixing this now. I haven't heard anyone complaining about this and there are more urgent issues to address. If you add the updated texts, I'll merge the PR.

madhuvardan-CMI:
done

PR #582, opened 2019-02-15
APPDEV-1263 Use regular log tags

Bert-R:
Instead of using a free text, just use the name of the origin class as tag for log messages, as recommended by the Android documentation.


Sivateja-CMI:
Looks good. thanks.

PR #589, opened 2019-03-04
APPDEV-1233 fabric crash in password encryption in anroid 8

Sivateja-CMI:
-- Merge from latest develop branch.
-- Code to set default password encryption method in user override request success.


Sivateja-CMI:
Yes, you are correct. 
I could only think of having both the lines in one function. But we still need to call it in all cases. right?
Alternately we can handle this just like we did for user entity by defining constant of this Enum rather setting it after registration.

Bert-R:
Let's leave it this way. I looked at it in the first PR of this fix and it requires quite a bit of refactoring to do this right.

PR #591, opened 2019-03-24
APPDEV-1269 Vpn and yona service to be stopped after unsubscribing.

Sivateja-CMI:
APPDEV-1269 Vpn and yona service to be stopped after un-subscribing.
 -- App now disconnects the VPN connection after successful un-subscribing at server end. Post to that app clears local cached user data and preferences.

### DIFF
--- a/app/src/main/java/nu/yona/app/YonaApplication.java
+++ b/app/src/main/java/nu/yona/app/YonaApplication.java
@@ -108,6 +108,9 @@ public class YonaApplication extends Application
 		return sharedAppDataState.getUser();
 	}
 
+	protected OpenVPNService vpnService;
+	private ServiceConnection connection;
+
 	@Override
 	public void onCreate()
 	{
@@ -210,38 +213,39 @@ public class YonaApplication extends Application
 		return mContext.tracker;
 	}
 
-	protected OpenVPNService mService;
 
-	private final ServiceConnection mConnection = new ServiceConnection()
+	private ServiceConnection getVpnServiceConnection()
 	{
-
-		@Override
-		public void onServiceConnected(ComponentName className,
-									   IBinder service)
+		return connection = new ServiceConnection()
 		{
-			// We've bound to LocalService, cast the IBinder and get LocalService instance
-			OpenVPNService.LocalBinder binder = (OpenVPNService.LocalBinder) service;
-			mService = binder.getService();
-			stopVPN();
-			getAppContext().unbindService(mConnection);
-		}
+			@Override
+			public void onServiceConnected(ComponentName className,
+										   IBinder service)
+			{
+				// We've bound to LocalService, cast the IBinder and get LocalService instance
+				OpenVPNService.LocalBinder binder = (OpenVPNService.LocalBinder) service;
+				vpnService = binder.getService();
+				stopVPN();
+				getAppContext().unbindService(connection);
+			}
 
-		@Override
-		public void onServiceDisconnected(ComponentName arg0)
-		{
-			mService = null;
-			// TODO: Need to show user a alert on failure to stop vpn.
-		}
+			@Override
+			public void onServiceDisconnected(ComponentName arg0)
+			{
+				vpnService = null;
+				// TODO: Need to show user a alert on failure to stop vpn.
+				nu.yona.app.utils.Logger.loge(YonaApplication.class, "Error in binding the vpn service with " + arg0.getClassName());
+			}
 
-	};
-
+		};
+	}
 
 	private void stopVPN()
 	{
 		ProfileManager.setConntectedVpnProfileDisconnected(YonaActivity.getActivity());
-		if (mService != null && mService.getManagement() != null)
+		if (vpnService != null && vpnService.getManagement() != null)
 		{
-			mService.getManagement().stopVPN(false);
+			vpnService.getManagement().stopVPN(false);
 		}
 	}
 
@@ -249,7 +253,7 @@ public class YonaApplication extends Application
 	{
 		Intent intent = new Intent(YonaActivity.getActivity(), OpenVPNService.class);
 		intent.setAction(OpenVPNService.START_SERVICE);
-		getAppContext().bindService(intent, mConnection, Context.BIND_AUTO_CREATE);
+		getAppContext().bindService(intent, getVpnServiceConnection(), Context.BIND_AUTO_CREATE);
 	}
 
 	public void stopOpenVPNService()
--- a/app/src/main/java/nu/yona/app/YonaApplication.java
+++ b/app/src/main/java/nu/yona/app/YonaApplication.java
@@ -108,9 +108,6 @@ public class YonaApplication extends Application
 		return sharedAppDataState.getUser();
 	}
 
-	protected OpenVPNService vpnService;
-	private ServiceConnection connection;
-
 	@Override
 	public void onCreate()
 	{
@@ -216,7 +213,7 @@ public class YonaApplication extends Application
 
 	private ServiceConnection getVpnServiceConnection()
 	{
-		return connection = new ServiceConnection()
+		return new ServiceConnection()
 		{
 			@Override
 			public void onServiceConnected(ComponentName className,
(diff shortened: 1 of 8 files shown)


app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +209,53 @@ public static Tracker getTracker()
...
+			mService = binder.getService();
+			stopVPN();
+			getAppContext().unbindService(mConnection);
+		}
+
+		@Override
+		public void onServiceDisconnected(ComponentName arg0)
+		{
+			mService = null;
+			// TODO: Need to show user a alert on failure to stop vpn.
```
Bert-R: At least log it, so it can be analyzed.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +209,53 @@ public static Tracker getTracker()
 		}
 		return mContext.tracker;
 	}
+
+	protected OpenVPNService mService;
```
Bert-R: Declare this member along with the others at the top of the class. Also strip the ``m`` prefix, as that's not a convention we apply.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +209,53 @@ public static Tracker getTracker()
...
+
+	private final ServiceConnection mConnection = new ServiceConnection()
```
Bert-R: Same here. Next, initialize it through a call to a private method rather than putting all of this inline here. Also strip the ``m`` prefix, as that's not a convention we apply.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +209,53 @@ public static Tracker getTracker()
...
+	private void bindOpenVPNService()
+	{
+		Intent intent = new Intent(YonaActivity.getActivity(), OpenVPNService.class);
+		intent.setAction(OpenVPNService.START_SERVICE);
+		getAppContext().bindService(intent, mConnection, Context.BIND_AUTO_CREATE);
+	}
+
+	public void stopOpenVPNService()
+	{
+		bindOpenVPNService();
```
Bert-R: Inline that method here, as it is pointless on its own.

app/src/main/java/nu/yona/app/ui/settings/SettingsFragment.java
```diff
@@ -131,7 +135,8 @@ private View configureSettingsListView(View settingsLayoutView)
 				getString(R.string.adddevice),
 				getString(R.string.showopenvpnlog),
 				getString(R.string.contact_us),
-				getString(R.string.deleteuser)};
+				getString(R.string.deleteuser),
+				"STOP VPN", "START VPN"};
```
Bert-R: Why is this different from the rest of the list? The strings are now repeated here and below, causing a risk that they grow apart.

app/src/main/java/nu/yona/app/ui/settings/SettingsFragment.java
```diff
@@ -248,29 +261,37 @@ public void onClick(DialogInterface dialog, int which)
...
-			public void onError(Object errorMessage)
-			{
-				YonaActivity.getActivity().dismissLoadingView();
-				Snackbar snackbar = Snackbar.make(YonaActivity.getActivity().findViewById(android.R.id.content), ((ErrorMessage) errorMessage).getMessage(), Snackbar.LENGTH_SHORT);
-				TextView textView = ((TextView) snackbar.getView().findViewById(android.support.design.R.id.snackbar_text));
-				textView.setMaxLines(5);
-				snackbar.show();
-			}
-		});
+	private void finishUnSubscribe()
```
Bert-R: Unsubscribe is one word, so spell it with a lowercase ``s``

app/src/main/java/nu/yona/app/ui/LaunchActivity.java
```diff
@@ -115,7 +115,7 @@ else if (!getSharedUserPreferences().getBoolean(PreferenceConstant.STEP_PASSCODE
 			bundle.putInt(AppConstant.COLOR_CODE, ContextCompat.getColor(LaunchActivity.this, R.color.grape));
 			startNewActivity(bundle, PasscodeActivity.class);
 		}
-		else if (!TextUtils.isEmpty(getSharedUserPreferences().getString(PreferenceConstant.YONA_PASSCODE, "")))
+		else if (!TextUtils.isEmpty(getSharedUserPreferences().getString(PreferenceConstant.YONA_DATA, "")))
```
Bert-R: Why is this?

Sivateja-CMI: There should not be much difference. But, once I faced issue after un-subscribing. Later I was unable to reproduce. I suspect some thing with cache and I am more sure 'YONA_DATA' is cleared from cache.

app/src/main/java/nu/yona/app/YonaApplication.java
Sivateja-CMI: bindOpenVPNService can be used in future, to do any other operations on service other than stopping.

Sivateja-CMI:
Yes, VPN will be connected only of the user entity is available.

app/src/main/java/nu/yona/app/YonaApplication.java
Bert-R: That's not the case. This method binds a handler that stops the VPN, so it should be named ``stopOpenVPNService``.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +212,54 @@ public static Tracker getTracker()
 		}
 		return mContext.tracker;
 	}
+
+
+	private ServiceConnection getVpnServiceConnection()
+	{
+		return connection = new ServiceConnection()
```
Bert-R: This is a weird construction. Why are you assigning that member here? You're only using it on line 229, right? Can't you simply use ``this`` there and do away with the member?

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -101,6 +108,9 @@ public static User getAppUser()
 		return sharedAppDataState.getUser();
 	}
 
+	protected OpenVPNService vpnService;
```
Bert-R: This can be private, right?

app/src/main/java/nu/yona/app/YonaApplication.java
Bert-R: This one isn't addressed.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +212,54 @@ public static Tracker getTracker()
...
+			{
+				// We've bound to LocalService, cast the IBinder and get LocalService instance
+				OpenVPNService.LocalBinder binder = (OpenVPNService.LocalBinder) service;
+				vpnService = binder.getService();
+				stopVPN();
+				getAppContext().unbindService(connection);
+			}
+
+			@Override
+			public void onServiceDisconnected(ComponentName arg0)
```
Bert-R: Use an appropriate name.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +212,54 @@ public static Tracker getTracker()
...
+			{
+				vpnService = null;
+				// TODO: Need to show user a alert on failure to stop vpn.
+				nu.yona.app.utils.Logger.loge(YonaApplication.class, "Error in binding the vpn service with " + arg0.getClassName());
+			}
+
+		};
+	}
+
+	private void stopVPN()
```
Bert-R: Make this part of the anonymous class and add the VPN service as a parameter, so you don't need a member variable for it.

app/src/main/java/nu/yona/app/YonaApplication.java
```diff
@@ -202,5 +212,54 @@ public static Tracker getTracker()
...
+	{
+		ProfileManager.setConntectedVpnProfileDisconnected(YonaActivity.getActivity());
+		if (vpnService != null && vpnService.getManagement() != null)
+		{
+			vpnService.getManagement().stopVPN(false);
+		}
+	}
+
+	private void bindOpenVPNService()
```
Bert-R: If you pass the connection as a parameter, you can retain this as a generally usable method.

app/src/main/res/values-nl/strings.xml
```diff
@@ -231,4 +231,6 @@
     <string name="yona_vpn_notification_channel_name">VPN-meldingen</string>
     <string name="mandatory_user_fetch_failure">De Yona-app heeft deze keer een netwerkverbinding nodig om te starten.</string>
     <string name="usage_access_unavailable_message">Sta Yona in het instellingenmenu toe om de gebruiksstatistieken van apps in te zien.</string>
+    <string name="stop_vpn_setting">HOU OP VPN</string>
```
Bert-R: The English texts are appropriate in Dutch too in this particular case.

app/src/main/java/nu/yona/app/ui/LaunchActivity.java
Sivateja-CMI: I suspect, this could be the reason for issue in the Vishal's phone, I am planning to create separate create a PR with extra logging around this and test in Vishal's phone.

Sivateja-CMI: @Bert-R , can I add the needed logging in this PR?

Bert-R: Yes, no problem

Bert-R:
@Sivateja-CMI Did you have a chance to thoroughly review and test the changes that I pushed?

Sivateja-CMI:
Yes, bert, I have tested the regular user cases, I need to test the device restart scenario, I will complete it as first thing today.

Sivateja-CMI:
Tested the app. It seems to work fine. I am ok with the changes.

PR #592, opened 2019-04-18
APPDEV-1274 signing up against the new server fails

madhuvardan-CMI:
-Handling of 504 response by displaying a message to the user

### DIFF
--- a/app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
@@ -126,7 +126,7 @@ class BaseImpl
 				.connectTimeout(NetworkConstant.API_CONNECT_TIMEOUT_IN_SECONDS, TimeUnit.SECONDS)
 				.writeTimeout(NetworkConstant.API_WRITE_TIMEOUT_IN_SECONDS, TimeUnit.SECONDS)
 				.readTimeout(NetworkConstant.API_READ_TIMEOUT_IN_SECONDS, TimeUnit.SECONDS)
-				.retryOnConnectionFailure(false)
+				//.retryOnConnectionFailure(false)
 				.addInterceptor(getInterceptor)
 				.build();
 	}
--- a/app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
+++ b/app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
@@ -46,41 +46,38 @@ class BaseImpl
 {
 	private final int maxStale = 60 * 60 * 24 * 28; // keep cache for 28 days.
 	//    public final String localLanguage = ;
-	private final Interceptor getInterceptor = new Interceptor()
-	{
-		@Override
-		public Response intercept(Chain chain) throws IOException
-		{
-			Response response = null;
-			Request request = chain.request();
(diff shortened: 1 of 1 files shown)


app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
```diff
@@ -46,42 +46,38 @@
...
-					.header("Cache-Control", "public, max-age=" + maxStale)
-					.build();
+			chain.request().newBuilder().addHeader("Cache-Control", "only-if-cached").build();
+		}
+		else
+		{
+			throw new IllegalStateException(YonaApplication.getAppContext().getString(R.string.server_not_rechable));
 		}
+		chain.request().newBuilder().addHeader(NetworkConstant.CONTENT_TYPE, "application/json");
+		Response response = verifyResponse(request, chain);
```
Bert-R: Change this into:
```
		Response response = chain.proceed(request);
		verifyResponse(response, chain);
```

app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
```diff
@@ -46,42 +46,38 @@
...
+		Response response = chain.proceed(request);
+		if (response.priorResponse() != null && response.priorResponse().code() == HttpURLConnection.HTTP_MOVED_PERM)
+		{
+			throw new IllegalStateException(YonaApplication.getAppContext().getString(R.string.server_not_rechable));
+		}
+		else if (response.code() == HttpURLConnection.HTTP_GATEWAY_TIMEOUT)
+		{
+			throw new SocketTimeoutException(YonaApplication.getAppContext().getString(R.string.server_not_rechable));
+		}
+		request.newBuilder().build();
```
Bert-R: What does this line do?

app/src/main/java/nu/yona/app/api/manager/network/BaseImpl.java
```diff
@@ -46,42 +46,38 @@
...
 	};
 
+	private Response verifyResponse(Request request, Interceptor.Chain chain) throws IOException
+	{
+		Response response = chain.proceed(request);
+		if (response.priorResponse() != null && response.priorResponse().code() == HttpURLConnection.HTTP_MOVED_PERM)
+		{
+			throw new IllegalStateException(YonaApplication.getAppContext().getString(R.string.server_not_rechable));
+		}
+		else if (response.code() == HttpURLConnection.HTTP_GATEWAY_TIMEOUT)
```
Bert-R: There are way more error codes that don't return JSON, so the check should start with:
```
if (Family.familyOf(response.code()) != Family.SUCCESS)
```
Then you can make an attempt to interpret the response to see whether it's JSON you understand, and if not, just throw an exception that includes the status code. Otherwise, you can take the error from the interpretation of the JSON.

Sivateja-CMI: In some cases we are handling 400's in the callbacks ( for e.g user override), this will disturb that logic

Bert-R: Yes, I understand. That's why I talk about "attempt to...". In the server, we're also making rest requests. You might want to look into [RestClientErrorHandler.java](https://github.com/yonadev/yona-server/blob/master/core/src/main/java/nu/yona/server/rest/RestClientErrorHandler.java) to see how we "attempt to..." there.

